PR #4, opened 2016-03-19
Badges - initial implementation

hakonrossebo:
This is an incomplete implementation. I just want to check if I'm on track and get some feedback.

- Currently, I've only  implemented badges for Icons. If this implementation is feasible, I need some input on how to add this info on other element types like Button - can we change the Button model to be a record containing  the current Ripple + BadgeInfo? 
- As I'm just starting out with Elm/functional programming, I appreciate feedback on how to improve. I'm sure mye coding style is pretty elementary now.


debois:
Hi Håkon, 

Great! Much appreciated. I love it that you added documentation already.

The problem you're solving with BadgeInfo and adding a parameter to Icon.view is that you need to apply additional CSS styling, yes? I think this is a good idea, but it can be done a bit more smoothly by adding not a BadgeInfo to Icon.view, but rather a Style from the Material.Style module. Ideally, you'd change Icon.view to have the type

    Icon.view : String -> Style -> Html 

Then express both the current Size parameter and BadgeInfo as values of type Style. Would you be interested in submitting a new pull request along these lines?

debois:
Since you solicited comments, I thought I'd make a few:

- To me, your work looks pretty much like idiomatic Elm. (Say, the introduction of Icon.viewWithOptions.)
- Usually, I'd expect a view function (at least in the Elm Architecture) to have the type 
    
        view : ... -> Model -> Html 

Badge.view should probably have had a different name. (But I'm proposing it should become a value of type Style instead, see above). 

- Badge.view is neat and idiomatic. You could consider using String.join instead of adding spaces yourself and using the (otherwise very  neat) `List.fold (++) ""` 

Altogether, I like your Elm-style. It looks fine to me :)

hakonrossebo:
Thanks. Good to hear that I'm not on the wrong path. This is a great way to learn. 

I fetched the updates and I'm unable to locate/compile Material.Style - is it missing from the last commits?

debois:
Sorry. Try again, please.

hakonrossebo:
I really like the Style implementation as a way to get around the limitations on existing attributes. The Badge implementation also require a separate data-badge attribute. Should I implement this attribute functionality in Material.Style and append to attrs? It may not be a good match with the Style name?

I can change the Icon implementation to use Style if you have not done so already.

debois:
Thanks. Probably adding a Style case producing custom attributes is a good way to go. You're right it doesn't sit perfectly with the name "Style", though.  I'm not sure what would be a better name—Config? Setup? Attributes? (Not Attributes; the point of Style was that it enables CSS styling, by adding classes if necessary. We don't want to end up re-implementing HTML.Attributes.)

Maybe stick with Style for now, and add the custom-attribute case for badges? You can kind of make the argument that badges has to do with styling. Later well see if we need it for other things; then we'll know what we call Styke is really an abstraction of :)

debois:
Oh, and it'd be great if you added Style to Icon. :)

hakonrossebo:
I'v updated the Badge implementation to use Style. I'll create a separate PR to update Icon.elm. The Badge Demo also needs more examples, but that can probably also be a separate PR. The Badge Demo only styles span now, but I'll update with Icon when the implementation is ready.

hakonrossebo:
Forgot to revert Icon.elm back to initial implementation. Updated PR with initial implementation.

debois:
Great! 

Would it make sense to split the [badgeStyle](https://github.com/debois/elm-mdl/pull/4/files#diff-8c9ffc88f58000de3a4235a059ce4ff6R66) function into three distinct Style properties? It would require adding a fifth constructor to Style:

    type Style
      = Class String
      | CSS (String, String)
      | Attr (String, String)
      | Multiple (List Style)
      | NOP

Then Badge could get the very neat interface:

    module Badge where 
    ...
    withBadge : String -> Style  -- Maybe there's a better name. 
    noBackground : Style
    overlap : Style 

?

hakonrossebo:
Elegant way to solve Badge styling. I updated the PC - think I understood, but I could need some guidance on how to map out the Multiple style type in Styles.styled. Possible that I'm making this too complex :)

debois:
I think you're doing it just right. You need a recursive function since Style is now a recursive type:

    flatten : Style -> List Style -> List Style
    flatten style styles = 
      case style of
        Multiple styles' ->
          foldl flatten styles' styles
        style ->
          style :: styles

Then Styles.styled, preprocess `styles` into something without Multiples by doing

    let
       flat = foldl flatten [] styles
       ...

hakonrossebo:
Thanks, updated the PR with a working implementation now.

hakonrossebo:
Cleaned up some comments and added HTML symbols to example

debois:
Very good, thanks!

Issue #13, opened 2016-03-23
Layout keyboard shortcuts


debois:
The drawer in Layout should open and close on various keyboard shortcuts. Specifically:

 - [Close on escape](https://github.com/google/material-design-lite/blob/master/src/layout/layout.js#L168-L174) if open and not fixed.
 - [Open and close on enter/spacebar](https://github.com/google/material-design-lite/blob/master/src/layout/layout.js#L200-L210) when the drawer button is selected.

PR #15, opened 2016-03-24
Icon changed to implement Style

hakonrossebo:
I've updated the Icon to use Style. Not sure if it is better to have separate Icon Size Style functions - tried a single function now.

As for Icon sizing, I'm not able to change the actual icon size, but I was not able to do that earlier either.


debois:
Cool. I think your right that it would be a good idea to replace iconSize with separate style values, like this

    size18 : Style
    size24 : Style
    ...

By the way, since people will like be using the library qualified (doing `import Material.Icon as Icon`, then writing `Icon.size`), I don't think we need to put "Icon" in the names of things in Material.Icon—see also [this link](http://package.elm-lang.org/help/design-guidelines#module-names-should-not-reappear-in-function-names).

hakonrossebo:
Updated PR now. Seems better with separate values. Agree that module names should not appear in function names.

debois:
Neat! 

Where do you want to go next? The Card and List components has no JavaScript and are mostly a matter of finding the right API. You could do them exclusively as Styles, or you could come up with a more precise API—I think the latter is likely the better option. I'd be happy to discuss what such an API should look like—if you want to go this way, maybe submit a PR with only API (no actual implementation) as a starting point?

Otherwise, you could look at some of the issues I've opened. Maybe [13](https://github.com/debois/elm-mdl/issues/13) is a good starting point?

hakonrossebo:
I think I will just start by looking into how Cards and Lists work in MDL and how you have implemented the rest of the API first :)

debois:
Excellent. Could you open an issue or submit a pull request so that we have some place to discuss details?

debois:
I'm playing around with Github a bit. Would you mind if I opened an issue for Cards (or Lists) and assigned you to do it?

Also, feel free to ask if you have any questions.

hakonrossebo:
Yes, please assign issues to me :)

Waffle.io is also a great way to keep track of issues.

debois:
I’ve assigned you :)

Issue #7, opened 2016-03-21
Router integration for demo tabs


debois:
Each tab [in the demo](https://debois.github.io/elm-mdl/wip.html) should have its own URL so that they can be linked directly. E.g., going to https://debois.github.io/elm-mdl/#grid should go to the demo with the Grid tab selected. 

Elm has [various routers](https://groups.google.com/forum/#!topic/elm-discuss/e2KlXOU3nkM) one might integrate with.

PR #28, opened 2016-04-05
Add Hop router to Demo Example.

groob:
Hi @debois, 

I added the sporto/hop router to the demo module to address #7.
With this addition the tabs are activated by visiting the url (example `http://localhost:8000/#grid`) or by clicking the tab in the UI. When the tab is clicked in the UI, the path is also updated. 

Currently the routes are hardcoded in the `changeTab` function and also in `Routing.matchers`. A bit ugly, but perhaps it's sufficient for the demo. What do you think?

### DIFF
--- a/examples/Demo.elm
+++ b/examples/Demo.elm
@@ -1,3 +1,5 @@
+module Main (..) where
+
 import StartApp
 import Html exposing (..)
 import Html.Attributes exposing (href, class, style)
@@ -7,46 +9,54 @@ import Task
 import Signal
 import Task exposing (Task)
 import Array exposing (Array)
-
+import Routing
+import Hop
+import Hop.Navigate exposing (navigateTo)
 import Material.Color as Color
+import Material.Layout
 import Material.Layout as Layout exposing (defaultLayoutModel)
 import Material exposing (lift, lift')
 import Material.Style as Style
-
 import Demo.Buttons
 import Demo.Grid
 import Demo.Textfields
 import Demo.Snackbar
 import Demo.Badges
---import Demo.Template
 
+
+--import Demo.Template
 -- MODEL
 
 
 layoutModel : Layout.Model
 layoutModel =
   { defaultLayoutModel
-  | state = Layout.initState (List.length tabs)
+    | state = Layout.initState (List.length tabs)
   }
 
 
 type alias Model =
   { layout : Layout.Model
+  , routing : Routing.Model
   , buttons : Demo.Buttons.Model
   , textfields : Demo.Textfields.Model
-  , snackbar : Demo.Snackbar.Model
-  --, template : Demo.Template.Model
+  , snackbar :
+      Demo.Snackbar.Model
+      --, template : Demo.Template.Model
   }
 
 
 model : Model
 model =
   { layout = layoutModel
+  , routing = Routing.initialModel
   , buttons = Demo.Buttons.model
   , textfields = Demo.Textfields.model
-  , snackbar = Demo.Snackbar.model
-  --, template = Demo.Template.model
+  , snackbar =
+      Demo.Snackbar.model
+      --, template = Demo.Template.model
   }
+
 
 
 -- ACTION, UPDATE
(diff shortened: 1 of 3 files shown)


examples/Demo.elm
```diff
@@ -57,37 +67,106 @@ type Action
...
+
+        model' =
+          { model | routing = routing' }
+      in
+        ( model'
+        , Effects.map RoutingAction fx
+        )
+
+    HopAction _ ->
+      ( model, Effects.none )
```
debois: It looks weird that the HopAction doesn't do anything. Is it because it's really the HopAction in the Router module that counts?

debois:
... but it's **awesome** that you did this!

I think you're right that it'd be much nicer if the routes were a part of the `tabs` value in [Demo.elm](https://github.com/debois/elm-mdl/blob/master/examples/Demo.elm#L109-L123). Is that a change you'd be willing to do?

(I'll make a few more line notes in a few moments.)

But, awesome! I'm so happy you did this!

examples/Demo.elm
```diff
@@ -143,66 +233,111 @@ stylesheet = Style.stylesheet """
...
-            , ("padding-left", "5%")
-            , ("padding-right", "5%")
+  routingView addr model
+
+
+routingView : Signal.Address Action -> Model -> Html
+routingView addr model =
+  case (Debug.log "Route " model.routing.route) of
+    Routing.Home ->
+      let
```
debois: I don't think the `let` here adds clarity—Elm is verbose enough as it is :). Maybe just a one-liner:

    Routing.Home -> 
      appView addr { model | layout = setTab model.layout 0 }

Same for the next case, `Routing.TabRoute`.

examples/Demo.elm
```diff
@@ -143,66 +233,111 @@ stylesheet = Style.stylesheet """
...
+      your html, as done with page.html. Removing it will then
+      fix the flicker you see on load.
+      -}
+      |>
+        Material.topWithScheme Color.Teal Color.Red
+
+
+routerSignal : Signal Action
+routerSignal =
+  Signal.map RoutingAction Routing.signal
```
debois: Maybe inline this in the `input` value below?

examples/Demo.elm
```diff
@@ -57,37 +67,106 @@ type Action
...
+      lift .buttons (\m x -> { m | buttons = x }) ButtonsAction Demo.Buttons.update a model
 
+    TextfieldAction a ->
+      lift' .textfields (\m x -> { m | textfields = x }) Demo.Textfields.update a model
 
-type alias Addr = Signal.Address Action
+    SnackbarAction a ->
+      lift .snackbar (\m x -> { m | snackbar = x }) SnackbarAction Demo.Snackbar.update a model
 
+    RoutingAction a ->
```
debois: I think this case could become simply

    RoutingAction a -> 
      lift .routing (\m x -> { m | routing = x}) RoutingAction Routing.update a model

groob:
Thanks for the code review. I will update with the changes you suggested. 

`HopAction ()` It must be there and triggers the location change in the url. 
It's IMO not very readable, but it's explained in the [elm-tutorial](http://www.elm-tutorial.org/070_routing/navigation_from_list_to_edit.html) guide.  

> `Hop.Navigate.navigateTo path` returns an effect. When this effect is run by Elm the location of the browser will change. [...]
> We also have to handle `HopAction` because we are wrapping the Hop effect with it. After the location changes we will receive a `HopAction` with a payload, but we don't need to do anything with this here.

debois:
Ok, thanks! Looking forward to the updated every PR :)

groob:
I addressed the style comments. 

I started working on combining the tabs and route matchers to have a generic solution but that might be too complex and probably unrealistic to handle in a simple example like here.` 

Here is what I have so far:

```Elm

type alias TabAndRoute =
  { tabRoutes : List (PathMatcher Route)
  , tab : ( String, Addr -> Model -> List Html )
  }

tabsAndRoutes : List TabAndRoute
tabsAndRoutes =
  let
    snackbar =
      { tabRoutes =
          [ match1 Routing.Home "/"
          , match1 (Routing.TabRoute 0) "/snackbar"
          ]
      , tab =
          ( "Snackbar"
          , \addr model ->
              [ Demo.Snackbar.view (Signal.forwardTo addr SnackbarAction) model.snackbar ]
          )
      }

    textfields =
      { tabRoutes = [ match1 (Routing.TabRoute 1) "/textfields" ]
      , tab =
          ( "textfields"
          , \addr model ->
              [ Demo.Textfields.view (Signal.forwardTo addr TextfieldAction) model.textfields ]
          )
      }
  in
    [ snackbar, textfields ]


matchers : List (PathMatcher Route)
matchers =
  List.concatMap .tabRoutes tabsAndRoutes


tabs : List ( String, Addr -> Model -> List Html )
tabs =
  List.map .tab tabsAndRoutes

```
The next step would be to figure out how to get the tab position and path from the `tabsAndRoutes` list and to pass that to the `update`.

debois:
I like the approach!

Maybe it would be better if `tabs` looked like this:

    tabs =
      [ ("Snackbar"  
        ,  \addr model ->
             [Demo.Snackbar.view (Signal.forwardTo addr SnackbarAction) model.snackbar])
        , "/snackbar"
        )
      , ...
      ]

Then use List.indexedMap with a function something like

    \idx (_, _, path) -> 
      [ match1 (Routing.TabRoute idx) path ]

(You'd need to special-case it for idx == 1, it seems.)

Is this something you'd be interested in doing, or should I take it from here?

groob:
I'm trying to work on a separate project that uses both mdl and hop. 
I might get a better idea of how to match Layout with routes then. 

If you'd like to pick it up from here, go ahead. I'll contribute if I think of something too.

debois:
Ok, cool. Appreciate the contribution!

debois:
Just in case you're interested: I inlined the Router module and computed path marchers directly from the tabs value based on your starting point above.

If you need anything from the library let me know—I need a few more things myself, and I'd be happy to help out with others.

PR #32, opened 2016-04-15
Correct possible mistake in README

shamrin:


debois:
Well spotted, thanks!

Just in case you're interested: The README.md file is both the entry point for GitHub and the [elm-package repository](http://package.elm-lang.org/packages/debois/elm-mdl/latest/). It works well enough for GitHub, but I think its confusing on packages.elm-lang.org. Do you want to take a shot at editing it so it fits both?

PR #33, opened 2016-04-16
Menus

aforemny:
Hi, I made an attempt at implementing menus! :)

It is still work-in-progress, but I wanted to share it. Any chance of getting this merged when it's done?
Note that I am using native code to query a few javascript properties of the menu element on button press. I think I cannot do this in pure Elm, can I?

Demo: https://aforemny.github.io/elm-mdl/#/menus


debois:
Hi Alexander, 

Awesome! I definitely want to merge this when it's done!

About native code: We can't have native code in elm-mdl, because then we don't get to publish it as a package. But this is not a showstopper, I thin, because you *can* read Javascript properties off the DOM in pure Elm: you construct a `Json.Decode`r that traverses the DOM event you get and collect the information you need. 

I'm doing a similar thing in [the Ripple animation implementation](https://github.com/debois/elm-mdl/blob/master/src/Material/Ripple.elm#L118-L123). Since you need to call `getClientBoundingRect`, you probably need to use [this library](http://package.elm-lang.org/packages/debois/elm-dom/1.0.0/DOM) (ripple animations also use it). Let me know if you get problems with this, I'll be very happy to assist. 

Thanks!

Søren

debois:
I opened Issue #33 for this, just to signal to other's that the Menu component is already being worked on.

aforemny:
Hi!

Thanks for your advice. It made me question my assumption that DOM traversal in `Decoder a` is somehow limited to `e.target`. Turns out I /can/ decode 
> target.nextSibling.childNodes[1].childNodes.

However, `childNodes` is a `NodeList` (which is not an array), so I still need the following native code to traverse the nodeList:

```
 var decodeNodeList = function(decoder) {
    return function(value) {
      if (value instanceof NodeList) { // basically only this line changed
        var len = value.length;
        var list = List.Nil;
        for (var i = len; i--; ){
          list = List.Cons( decoder(value[i]), list );
        }
        return list;
      }
      crash('a NodeList', value);
    };
  };
```

I reckon that `decodeNodeList` could be included in core… I will try to adapt my code on Monday.

PS. I forgot to include `Material.Menu.Oracle` in my original PR. I uploaded it for completeness' sake, but it is about to change on Monday.

debois:
Hi Alexander, 

Very cool!

If you can decode 

    target.nextSibling.childNodes[1].childNodes

couldn't you then traverse `childNodes` using a decoder which repeatedly tries to decode 

    target.nextSibling.childNodes[1].childNodes[i]

for `i=0` and up until the decoder fails?

I'm really keen to have the menu component not rely on native because elm-mdl with a native component won't be allowed in the [package repository](http://package.elm-lang.org/).

About getting your `nodeList` decoder into Core. I don't think this will happen in the short term, for two reasons: (1) I think Core is de facto not looking at PRs while 0.17 is being worked on, and (2) your decoder opens up the broader question of what would actually be the best way for an Elm program to access the DOM. I could imagine some people saying that using JSON decoders the way we are is a hack, and there should be some more principled way, possibly involving a language extension, or simply that we should "wait for 0.17". (I could be wrong about both, though; maybe ask on #core in the elm-slack?)

At any rate, I don't want to wait for any of those things: I need Elm to be useful to me __now__. So as an interrim solution, can I propose you construct a `nodeList` decoder using not native but the above hack, then submit a PR for [elm-dom](https://github.com/debois/elm-dom/blob/master/src/DOM.elm)?

Looking forward to see more stuff monday!

Søren

src/Material/Menu/Oracle.elm
```diff
@@ -0,0 +1,123 @@
...
+        , offsetHeight = offsetHeight
+        , bounds = bounds
+        }
+    )
+    DOM.offsetTop
+    DOM.offsetLeft
+    DOM.offsetHeight
+    boundingClientRect
+
+boundingClientRect =
```
debois: Does this work!?! Do you actually get to call getBoundingClientRect like this?

src/Material/Menu/Oracle.elm
```diff
@@ -0,0 +1,123 @@
...
+      , bottom = 0
+      , right = 0
+      , width = width
+      , height = height
+      })
+    (position 0 0)
+    DOM.offsetWidth
+    DOM.offsetHeight
+
+position : Float -> Float -> Json.Decoder (Float, Float)
```
debois: Help me understand what happens here: It looks like you found a way to approximate getBoundingClientRect without traversing the DOM? How does this work?

aforemny:
Hello,

I just wanted to let you know that I uploaded some bogus Oracle.elm file. Oops. That line that you quoted, in particular `childNodes[1]` does not work, and I still need Native to decode `childNodes`. Sorry for the confusion.

debois:
Ok. I'all try to take a look at decoding childNodes it tonight or tomorrow.
The menu itself (with Native for now) is still work in progress, yes?

tir. 19. apr. 2016 kl. 01.10 skrev Alexander Foremny <
notifications@github.com>:

> Hello,
>
> I just wanted to let you know that I uploaded some bogus Oracle.elm file.
> Oops. That line that you quoted, in particular childNodes[1] does not
> work, and I still need Native to decode childNodes. Sorry for the
> confusion.
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/debois/elm-mdl/pull/33#issuecomment-211625295>
>
-- 
Søren Debois
+45 5358 5818

aforemny:
Yes, it is still work in progress. I will refactor this PR to use Native as we have discussed. I will manage to in the next couple of days. Yesterday something came up. :)

debois:
Ok, cool :)

I found a way to decode `childNodes` and added it to elm-dom. If you update to `1.1.0`, you'll find 
[this decoder](https://github.com/debois/elm-dom/blob/1.1.0/src/DOM.elm#L88-L101). (I exploited that `childNodes` apparently have fields `0`, `1`, ... for its elements.)

Was that the last obstacle to getting rid of Native?

aforemny:
Wow! That's pretty neat! That should solve the need for Native. I will test. :)

debois:
Excellent—looking forward to seeing the results!

aforemny:
This is a working implementation. It differs from mdl-menus in at least the following aspects:

- The hover effect on the button for some reason stays until you click.
- The menu does not close when clicked outside.
- I only implemented mouse input.

I did not really get the component model yet, so I removed types and functions related to it.

Can we merge this? In that case I could make it a single commit and open a new PR.

Demo: https://aforemny.github.io/elm-mdl/#/menus

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+
+{-| Model of the menu; common to all kinds of menus.
+Use `model` to initialise it.
+-}
+
+type alias Model =
+  { alignment : Style
+  , ripple : Bool
+  , items : Dict Int Ripple.Model
+  , open : Maybe Bool --
```
debois: I'm unsure about the two fields `open : Maybe Bool` (why Maybe?) and `closing : Bool`. Looking at the later code, I can't help wondering if it'd be better to encode the possible states directly:
```
type AnimationState 
  = Idle
  | Opening
  | Open
  | Closing
``` 
Would this make the update function clearer?

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+
+-- MODEL
+
+
+{-| Model of the menu; common to all kinds of menus.
+Use `model` to initialise it.
+-}
+
+type alias Model =
+  { alignment : Style
```
debois: As far as I can tell, the alignment doesn't need to live in the Model; it is only used in view. Could we take it out of the model and make it a parameter to view instead?

(I was in doubt myself what's the best way to do this for the longest time. I'm convinced by now that in general, you want to take stuff out of the model and put it as argument to view if you can get away with it. The existing components don't always do this yet, but they will eventually.)

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+  , ripple : Bool
+  , items : Dict Int Ripple.Model
+  , open : Maybe Bool --
+  , closing : Bool
+  , oracle : Maybe Oracle
```
debois: I think we need a better name than `oracle`. It fit perfectly when you got it with Native; now, not so much. It'd also be better if the extra module was somehow tied to menu. Maybe `Material.Menu.Geometry`?

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+  }
+
+
+{-| Component model styles.
+Used with `model`.
+-}
+
+bottomLeft : Style
+bottomLeft = cs "mdl-menu--bottom-left"
+
```
debois: Two blank lines between top-level definitions. See the [style guide](http://elm-lang.org/docs/style-guide#declarations).

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+type alias Item =
+  { html    : Html
+  , enabled : Bool
+  , divider : Bool
+  }
+
+{-| Item constructor. Pass `True` if there should be a divider below this item; `False` otherwise. Pass as second argument `True` if item should be enabled, and pass as third argument any `Html` as content of the menu item.
+-}
+
+item : Bool -> Bool -> Html -> Item
```
debois: I like it that an Item contains Html and not just text. 

When you are providing no defaults, I don't think it is necessary to provide a constructor function. (Don't Elm generate a function `Item : Html -> Bool -> Bool -> Item` automatically?)

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+  { html = html
+  , enabled = enabled
+  , divider = divider
+  }
+
+
+-- ACTION, UPDATE
+
+
+{-| Component action. The `Select n` action fires when the `n`th Item is selected; `Open`, `Close` when the menu opens, closes.
```
debois: It'd be good to know what `Close'` does. (There's no better name?)

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+-}
+
+-- Note: The Close' index is 1-based while the Select index is 0-based.
+
+type Action =
+    Toggle Oracle
+  | Close' Int Oracle
+  | Hide Int Oracle
+
+  | Select Int
```
debois: Would it be an idea to have `Item` contain the action to dispatch when the item is selected, rather than giving the `Select n` action? I'm not sure; I did something like that in [Snackbar](https://github.com/debois/elm-mdl/blob/master/src/Material/Snackbar.elm#L49-L71), but I'm not sure I liked how that turned out. 

(It makes it difficult to work with component support.)

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+
+{-| Component update.
+-}
+update : Action -> Model -> (Model, Effects Action)
+update action model =
+
+  case action of
+
+
+    -- If the menu is closed, we transition into opening state.
```
debois: That sounds really weird?

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+                    case model.open of
+                      Nothing   -> Just False
+                      Just True -> Nothing
+                      _         -> model.open
+
+                , oracle =
+                    case model.open of
+                      Nothing   -> Just oracle
+                      Just True -> Just oracle
+                      _         -> model.oracle
```
debois: I can't figure out what would make you keep the old oracle?

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+
+
+{-| Model of the menu; common to all kinds of menus.
+Use `model` to initialise it.
+-}
+
+type alias Model =
+  { alignment : Style
+  , ripple : Bool
+  , items : Dict Int Ripple.Model
```
debois: Consider hiding away the parts of the model you don't want users to tamper with in an opaque data type. See the [library design guide](http://package.elm-lang.org/help/design-guidelines#keep-tags-and-record-constructors-secret).

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+{-| Component update.
+-}
+update : Action -> Model -> (Model, Effects Action)
+update action model =
+
+  case action of
+
+
+    -- If the menu is closed, we transition into opening state.
+    Toggle oracle -> effect
```
debois: See [style guide](http://elm-lang.org/docs/style-guide) on case expressions (`->` should be followed by newline).

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+      { model | open = if model.open == Just False then
+                           Just True
+                         else
+                           model.open
+      }
+
+    -- Immediately close the menu.
+    Hide idx oracle -> effect
+
+      ( Effects.tick (\_ -> Select idx) )
```
debois: `always`

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+
+      { model | open = Nothing, closing = False, oracle = Just oracle }
+
+    -- Close the menu after some delay.
+    Close' idx oracle -> effect
+
+      ( Effects.task
```
debois: Use `Material.Helpers.delay`

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+        <| Task.succeed (Hide idx oracle))
+
+      { model | closing = True }
+
+
+    Ripple idx action ->
+
+      let
+
+        (model', effects) = Dict.get idx model.items
```
debois: Maybe try running your code through elm-format and see if you like the result?

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+                         |> Ripple.update action
+      in
+
+        effect
+
+          ( Effects.map (Ripple idx) effects )
+
+          { model | items = Dict.insert idx model' model.items }
+
+    Select _ -> pure model
```
debois: These do nothing?

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+-}
+view : Signal.Address Action -> Model -> List Item -> List Html
+view addr model items =
+  [ styled button
+      [ cs "mdl-button"
+      , cs "mdl-js-button"
+      , cs "mdl-button--icon"
+      , Style.attribute (onClick addr Oracle.decode Toggle)
+      ]
+      [ Html.i [ Html.class "material-icons" ] [ Html.text "more_vert" ]
```
debois: Use `Material.Icon.i`.

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+      ]
+  , styled div
+      [ cs "mdl-menu__container"
+      , cs "is-upgraded"
+      , cs' "is-visible" (model.open == Just True)
+
+      , css' "width"
+          ( case model.oracle of
```
debois: This function contains `case model.oracle of Just oracle -> f oracle Nothing -> "auto" a lot. Couldn't the commonalities be split out into a separate functions?

(Elm follows Haskell enough that it is usually clearer to make short seemingly trivial helper functions rather than writing everything out inline, I think.)

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+              Just oracle ->
+                oracle.menu.bounds.width
+                |> toString
```
debois: Define a helper function 
```
toPx : Float -> String
toPx  = toString >> flip (++) "px"
```

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+          ( case model.oracle of
+              Just oracle ->
+                oracle.menu.bounds.height
+                |> toString
+                |> flip (++) "px"
+              Nothing     -> "auto"
+          )
+          ( isJust model.oracle )
+
+      , flip (css' "top") (((model.alignment == bottomRight) || (model.alignment == bottomLeft)) && (isJust model.oracle)) <|
```
debois: This and the following looks to me like alignment ought to be a datatype, i.e., 
```
type Alignement 
  = BottomRight
  | BottomLeft 
  | TopRight
  | TopLeft
  | Unaligned
```
You can then use `case` to produce the `cs` you need.

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+        , model.alignment
+        , cs' "is-animating" (model.open == Just False)
+        , css' "clip"
+          (case model.oracle of
+
+             Nothing     -> "auto"
+             Just oracle ->
+
+               let
+                 rect x y w h =
```
debois:
```
rect x y w h = 
  [x,y,w,h]
  |> List.map toPx
  |> String.join " "
  |> (\coords -> "rect(" ++ coords ++ ")")
```

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+                 width  = oracle.menu.bounds.width
+                 height = oracle.menu.bounds.height
+               in
+                 if model.open == Just True then
+
+                     rect 0 width height 0
+
+                   else
+
+                     if model.alignment == bottomRight then
```
debois: Type for alignment as indicated above and case would look much better here.

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+                       else if model.alignment == topRight then
+                         rect height width height width
+                       else
+                         ""
+          )
+          (isJust model.oracle)
+        ]
+        ( let
+
+            makeItem n item =
```
debois: Lift this out of the let and into a separate function.

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+                transitionDuration =
+                  constant.transitionDurationSeconds *
+                  constant.transitionDurationFraction
+
+                offsetTop n =
+                  case model.oracle of
+                    Nothing -> 0
+                    Just oracle ->
+                      oracle.offsetTops
+                      |> List.drop (n-1) -- n is 1-based
```
debois: Would it be better if `offsetTops` and friends were `Array` or `Dict` instead?

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+                      |> List.head
+                      |> fromJust "Menu.view: offsetTop"
```
debois: I see what you are doing, but I really don't think `fromJust` is something we want. The promise of Elm is no crashes; if we happen to be wrong and the oracle __is__ missing an element, I'd rather have weird run-time behaviour than a crash.

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+                  case model.oracle of
+                    Nothing -> 0
+                    Just oracle -> oracle.menu.bounds.height
+
+                itemDelay =
+
+                  if model.alignment == topLeft ||
+                     model.alignment == topRight then
+
+                      (height - offsetTop n - offsetHeight n) / height * transitionDuration
```
debois: Neat.

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+                      ((offsetTop n) / height * transitionDuration)
+                      |> toString
+                      |> (flip (++) "s")
+              in
+
+                    styled li
+                    ( (if item.enabled then
+                          Style.attribute (onClick addr Oracle.decode' (Close' n))
+                        else
+                          Style.attribute (Html.attribute "disabled" "disabled"))
```
debois: Is there some reasons you don't use `Html.Attribute.disabled`? (I ask because I had trouble with it in the past and hope you have shown me a solution.)

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+                      :: cs' "mdl-js-ripple-effect" model.ripple
+                      :: cs' "mdl-menu__item--full-bleed-divider" item.divider
+                      :: Style.attribute (Html.property "tabindex" (string "-1"))
+                      :: Style.attribute (Ripple.downOn "mousedown" (Signal.forwardTo addr (Ripple n)))
+                      :: Style.attribute (Ripple.downOn "touchstart" (Signal.forwardTo addr (Ripple n)))
+                      :: Style.attribute (Ripple.upOn "mouseup" (Signal.forwardTo addr (Ripple n)))
+                      :: Style.attribute (Ripple.upOn "mouseleave" (Signal.forwardTo addr (Ripple n)))
+                      :: Style.attribute (Ripple.upOn "touchend" (Signal.forwardTo addr (Ripple n)))
+                      :: Style.attribute (Ripple.upOn "blur" (Signal.forwardTo addr (Ripple n)))
+                      :: []
```
debois: Use the usual `[  , , , , ]` style when you don't need to cons onto something (like here).

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+                    ( (if item.enabled then
+                          Style.attribute (onClick addr Oracle.decode' (Close' n))
+                        else
+                          Style.attribute (Html.attribute "disabled" "disabled"))
+                      :: cs "mdl-menu__item"
+                      :: css' "transition-delay" itemDelay (isJust model.open)
+                      :: cs' "mdl-js-ripple-effect" model.ripple
+                      :: cs' "mdl-menu__item--full-bleed-divider" item.divider
+                      :: Style.attribute (Html.property "tabindex" (string "-1"))
+                      :: Style.attribute (Ripple.downOn "mousedown" (Signal.forwardTo addr (Ripple n)))
```
debois: Is it correct that you can't use `Ripple.view` here because menus happen to not have the usual inner ripple-div for items?

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+                      :: Style.attribute (Ripple.downOn "touchstart" (Signal.forwardTo addr (Ripple n)))
+                      :: Style.attribute (Ripple.upOn "mouseup" (Signal.forwardTo addr (Ripple n)))
+                      :: Style.attribute (Ripple.upOn "mouseleave" (Signal.forwardTo addr (Ripple n)))
+                      :: Style.attribute (Ripple.upOn "touchend" (Signal.forwardTo addr (Ripple n)))
+                      :: Style.attribute (Ripple.upOn "blur" (Signal.forwardTo addr (Ripple n)))
+                      :: []
+                    )
+
+                    ( (::) item.html <|
```
debois: `( item.html :: `

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+
+                        if model.ripple then
+
+                            [ Ripple.view
```
debois: Wait, if you do call Ripple.view, do you still need to set all the "mouseup" etc. handlers above?

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+                          Style.attribute (onClick addr Oracle.decode' (Close' n))
+                        else
+                          Style.attribute (Html.attribute "disabled" "disabled"))
+                      :: cs "mdl-menu__item"
+                      :: css' "transition-delay" itemDelay (isJust model.open)
+                      :: cs' "mdl-js-ripple-effect" model.ripple
+                      :: cs' "mdl-menu__item--full-bleed-divider" item.divider
+                      :: Style.attribute (Html.property "tabindex" (string "-1"))
+                      :: Style.attribute (Ripple.downOn "mousedown" (Signal.forwardTo addr (Ripple n)))
+                      :: Style.attribute (Ripple.downOn "touchstart" (Signal.forwardTo addr (Ripple n)))
```
debois: Consider doing `let ripple = Signal.forwardTo addr (Ripple n))` someplace earlier.

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+onClick : Signal.Address Action -> Decoder Oracle -> (Oracle -> Action) -> Attribute
+onClick addr decoder action =
+  Html.onWithOptions
+  "click"
+  defaultOptions
+  decoder
+  (Signal.message addr << action)
+
+
+isJust : Maybe a -> Bool
```
debois: You'd have `isJust x == (x /= Nothing)`. I think it's easier to get rid of `isJust` and simply do `x /= Nothing` whenever you need it.

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+isJust x =
+  case x of
+    Nothing -> False
+    Just  _ -> True
+
+
+fromJust : String -> Maybe a -> a
+fromJust crash x =
+  case x of
+    Nothing -> Debug.crash crash
```
debois: See above comment on `Debug.crash`. I'd rather this function wasn't there.

src/Material/Menu/Oracle.elm
```diff
@@ -0,0 +1,87 @@
+module Material.Menu.Oracle
```
debois: This module is very neat.

src/Material/Menu/Oracle.elm
```diff
@@ -0,0 +1,87 @@
...
+  , offsetHeight : Float
+  , bounds : DOM.Rectangle
+  }
+
+{-| Decode Oracle from the button's reference
+-}
+
+decode : Decoder Oracle
+decode =
+  object5
```
debois: I think you can simply do `object5 Oracle`, where `Oracle` is an auto-generated version of the function you define in the next 7 lines.

src/Material/Menu/Oracle.elm
```diff
@@ -0,0 +1,87 @@
...
+    ( \button container menu offsetTops offsetHeights ->
+
+        { button = button
+        , menu   = menu
+        , container = container
+        , offsetTops = offsetTops
+        , offsetHeights = offsetHeights
+        }
+    )
+    ("target" := element)
```
debois: `(DOM.target element)` (I think.)

src/Material/Menu/Oracle.elm
```diff
@@ -0,0 +1,87 @@
...
+    (at ["target", "nextSibling"]  element)
```
debois: 1. Is it possible to get rid of the repeated `(at ["target", "nextSibling"] ...)`, maybe by splitting this decoder into two?
2. Could you add `nextSibling` to the `DOM` module and submit a PR? In that case, this could become the slightly neater `(DOM.target (DOM.nextSibling (DOM.childNode 1 element)))`. 

(If only Elm had a right-associate `<|`.)

src/Material/Menu/Oracle.elm
```diff
@@ -0,0 +1,87 @@
...
+    ( \button container menu offsetTops offsetHeights ->
+
+        { button = button
+        , container = container
+        , menu   = menu
+        , offsetTops = offsetTops
+        , offsetHeights = offsetHeights
+        }
+    )
+    (at ["target", "parentNode", "parentNode", "previousSibling"] element)
```
debois: Similar: Could you add `previousSibling` to DOM, then rephrase this in terms of that module?

src/Material/Menu/Oracle.elm
```diff
@@ -0,0 +1,87 @@
...
+    (at ["target", "parentNode"] (DOM.childNodes DOM.offsetTop))
+    (at ["target", "parentNode"] (DOM.childNodes DOM.offsetHeight))
+
+{-| Decode an Element
+-}
+
+element : Decoder Element
+element =
+  object4
+    (\offsetTop offsetLeft offsetHeight bounds ->
```
debois: `Element`

demo/Demo.elm
```diff
@@ -115,7 +119,7 @@ nth k xs =
 
 update : Action -> Model -> ( Model, Effects Action )
 update action model =
-  case Debug.log "Action" action of
```
debois: Thanks.

demo/Demo/Menus.elm
```diff
@@ -0,0 +1,222 @@
...
+
+
+view' : View -> List Style -> Html -> Signal.Address Menu.Action -> Menu.Model -> Html
+view' view coloring elem addr model =
+  view addr model coloring [elem]
+
+
+menus : List (Index, (Style, String))
+menus =
+  [ (Menu.bottomLeft, "Lower left")
```
debois: Maybe say in the Html someplace that "lower left" etc. tells you where the menu appears, not where the menu icon is supposed to sit. (I was confused about it when first seeing the demo.)

demo/Demo/Menus.elm
```diff
@@ -0,0 +1,222 @@
...
+  , (Menu.bottomRight, "Lower right")
+  , (Menu.topLeft, "Top left")
+  , (Menu.topRight, "Top right")
+  ]
+  |> List.indexedMap (\i r -> (i, r))
```
debois: `  |> List.indexedMap (,)`

demo/Demo/Menus.elm
```diff
@@ -0,0 +1,222 @@
...
+type Action
+  = Action Index Menu.Action
+
+
+type alias Model =
+  { menus : Dict.Dict Index Menu.Model
+  }
+
+
+update : Action -> Model -> (Model, Effects.Effects Action)
```
debois: `import Effects exposing (Effects)` to get rid of the weird `Effects.Effects`.

demo/Demo/Menus.elm
```diff
@@ -0,0 +1,222 @@
...
+    bar rightAlign =
+      let
+        align =
+          if rightAlign then ("right", "16px") else ("left", "16px")
+      in
+        div
+          [ class "bar"
+          , style
+            [ ("box-sizing", "border-box")
+            , ("background", "#3F51B5")
```
debois: Maybe use `Style.div` and `Color.accent` instead? (I'd like the demo to show in general how to use the library.)

demo/Demo/Menus.elm
```diff
@@ -0,0 +1,222 @@
...
+      [
+      ]
+
+  in
+
+    div
+    [ class "section"
+    ]
+    [ div
+        [ class "container mdl-shadow--2dp"
```
debois: Use `Elevation.e2` instead.

elm-package.json
```diff
@@ -1,5 +1,5 @@
 {
-    "version": "2.1.3",
+    "version": "2.1.4",
```
debois: :)

demo/Demo/Menus.elm
```diff
@@ -0,0 +1,222 @@
...
+
+type alias View =
+  Signal.Address Menu.Action -> Menu.Model -> List Style -> List Html -> Html
+
+
+type alias View' =
+  Signal.Address Menu.Action -> Menu.Model -> Html
+
+
+view' : View -> List Style -> Html -> Signal.Address Menu.Action -> Menu.Model -> Html
```
debois: Can't figure out how this helps?

demo/Demo/Menus.elm
```diff
@@ -0,0 +1,222 @@
...
+          ]
+          [ div
+              [ class "wrapper"
+              , style
+                [ ("position", "absolute")
+                , align
+                , ("box-sizing", "border-box")
+                ]
+              ]
+              ( Menu.view (Signal.forwardTo addr (Action idx)) model' items
```
debois: Very good! It's great that `items` is a parameter to view and not part of the model.

src/Material/Menu.elm
```diff
@@ -0,0 +1,523 @@
...
+unaligned = cs "mdl-menu--unaligned"
+
+
+-- ITEM
+
+
+{-| Item model.
+-}
+
+type alias Item =
```
debois: Should probably live under `-- VIEW`.

debois:
Altogether: Excellent! Very impressive, especially the Oracle module. I'm very happy you did this!

As you can see, I made some comments, some stylistic, some semantic. I'm hoping you'll take a look and act on the ones that make sense to you. 

You mentioned that: 

> The hover effect on the button for some reason stays until you click.

I don't see that in the demo? However, on Mac OS in both Chrome and Safari, the menu doesn't open on click, only on keyboard input (enter, space). 

> The menu does not close when clicked outside.

I see that. I wonder why not?

> I only implemented mouse input.

Ok. 

> Can we merge this? In that case I could make it a single commit and open a new PR.

Sure! I'd love it if you'd take a look at some of my other comments first, but otherwise, sure!

> edit: I think I understand the component parts a bit better. I'll try to update it tonight.

Ok, cool. If you can make it work, that's be great; otherwise, I can do it later.

aforemny:
Thanks a lot for your detailed feedback! I'm agreeing with all of it, and will update accordingly. :)

debois:
Excellent :)

aforemny:
I changed the action type quite a bit. I wanted to outline what's necessary for the module internally to function.

This implementation does not use the `Opening` animation state, and does not set the css class `is-animating` on opening as the mdl implementation does. I cannot see any difference in behaviour, so we might leave it out as it does not require us to have a tick function that changes from `Opening` to `Opened`. What's your opinion on this?

Because all signals carry `Geometry` I would like to replicate them, so that the component's public signals `Open`, `Close` and `Select Int` do not expose Geometry. Do you agree?

I found a fix for the Chrome bug by setting "pointer-events: none" on the menu's `i` tag. I did not test with `Material.Icon` yet.

Regarding the fact that the menu does not close when clicked outside, how would you handle that? mdl binds to the document's `click` handler and checks if the click happened outside the menu.

Can we similarly register to document events in Elm?

debois:
As I said above: Very neat!

About Opening/is-animating: at least on Chrome, when I try the Demo, the
menu items appear __before__ the background animation is complete. Properly
that's why you need the is-animating and opening state. (Compare with the
mdl implementation here <https://getmdl.io/components/#menus-section>.)

About hiding the Geometry part of Actions. I think exposing public actions
without them is a good idea. You'd forward the public actions from the
private ones as effects, yes?

Thank you for the Chrome fix—it works nicely now :)

About outside clicks: I don't know of a good way to register clicks with
`document`. (Quick googling didn't give me anything, your luck might be
better.) But the Layout component has the exact same functionality for its
Drawer; maybe look at what the CSS behind the layout obfuscator
<https://github.com/debois/elm-mdl/blob/master/src/Material/Layout.elm#L438-L447>
works
and see if you can achieve something similar?

Again, excellent work!


On Mon, Apr 25, 2016 at 12:50 PM Alexander Foremny <notifications@github.com>
wrote:

> I changed the action type quite a bit. I wanted to outline what's
> necessary for the module internally to function.
>
> This implementation does not use the Opening animation state, and does
> not set the css class is-animating on opening as the mdl implementation
> does. I cannot see any difference in behaviour, so we might leave it out as
> it does not require us to have a tick function that changes from Opening
> to Opened. What's your opinion on this?
>
> Because all signals carry Geometry I would like to replicate them, so
> that the component's public signals Open, Close and Select Int do not
> expose Geometry. Do you agree?
>
> I found a fix for the Chrome bug by setting "pointer-events: none" on the
> menu's i tag. I did not test with Material.Icon yet.
>
> Regarding the fact that the menu does not close when clicked outside, how
> would you handle that? mdl binds to the document's click handler and
> checks if the click happened outside the menu.
>
> Can we similarly register to document events in Elm?
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/debois/elm-mdl/pull/33#issuecomment-214262417>
>
-- 
Søren Debois
+45 5358 5818

aforemny:
Thank you for your detailed comments!

I spent some more time, updating the demo to use the component model, adding back the `Tick` action to menus to have the `is-animating` class set on opening, and hopefully most of your suggestions. :)

I would like to move on rather than working on this any more, so I merged this myself in feabf9f8100b76373c2a63b48a87806c765c46a6.

Issue #30, opened 2016-04-12
Textfield should implement also multi-line


debois:

debois:
Very vool, thanks! I added a few comments on potential improvements to the code below; let me know if you're interested in doing any.

PR #48, opened 2016-06-12
Add support for multi-line textfields (textareas)

vipentti:
Adds support for multi-line textfields using textareas to address #30 

Not 100% sure on using a string type' comparison to detect type for a textarea.

### DIFF
--- a/src/Material/Textfield.elm
+++ b/src/Material/Textfield.elm
@@ -3,6 +3,7 @@ module Material.Textfield exposing
   , onInput
   , Msg, Model, defaultModel, update, view
   , render
+  , textarea, rows, cols
   )
 
 {-| From the [Material Design Lite documentation](http://www.getmdl.io/components/#textfields-section):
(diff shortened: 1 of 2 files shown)


src/Material/Textfield.elm
```diff
@@ -236,6 +265,27 @@ view' lift model options =
       Options.collect defaultConfig options
     val = 
       config.value |> Maybe.withDefault model.value
+
+    isTextarea = config.type' == type' "textarea"
```
debois: Does this actually work? Even if it did, it might stop doing so in the future. I'd suggest instead introducing a type

    type Kind 
      = Input
      | TextArea
      | Password

Then change `Config.type` to have type `Kind` (maybe rename it `kind` in the process), then switch on `config.kind` in the appropriate places in `view`.

vipentti:
Thanks for the help! New to Elm so not quite sure yet about the most idiomatic way of doing things. 

Updated the Textfield to have an input Kind

PR #49, opened 2016-06-13
Add Transparent Header option to Layout

vipentti:
Current implementation of Transparency is not exclusive with other header modes.

Not sure whether to add a Transparent mode and have it be exclusive with the other modes or if being able to have for example a transparent scrolling header is intended. 

Also currently the example is using the background from the getmdl.io transparency example.


debois:
Very nice! I took the liberty of fancifying the demo part a little bit. 

Welcome to elm-mdl!

PR #50, opened 2016-06-14
Progress bars - Initial implementation

vipentti:
Initial implementation of MDL progress bars. Currently the bars are stateless and the progress and buffer values are updated using the properties. The type of the bar can also be set to `indeterminate`, this will use the `mdl-progress--indeterminate` class to animate the bar.
 

Sample use: 
```    
import Material.ProgressBar as ProgressBar

progressBar : Html m
progressBar = ProgressBar.view [ ProgressBar.progress 35, ProgressBar.buffer 85]
```

### DIFF
--- a/demo/Demo.elm
+++ b/demo/Demo.elm
@@ -27,6 +27,7 @@ import Demo.Snackbar
 import Demo.Badges
 import Demo.Elevation
 import Demo.Toggles
+import Demo.ProgressBar
 --import Demo.Template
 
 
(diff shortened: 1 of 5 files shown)


debois:
Bad luck: @aforemny concurrently implemented progressbars and spinners. Your demo is nicer, though. Would you be interested in (1) updating your demo to use Alexanders implementation, (2) add code samples from Alexenders Demo?

And, gentlemen, maybe we should begin opening issues saying "I'm implementing this" to avoid future duplication of effort?

Something else entirely. @vipentti: Could you look me up on the elm-slack (@debois)?

vipentti:
I can update the demo and add samples.

Probably easier to cancel this PR and make a new one for that ?

debois:
Either way is fine. @aforemny: Just so you know: @vipentti will be updating the demo part of "Loading" by integrating parts of this PR.

PR #57, opened 2016-06-15
Loading demo update

vipentti:
Update the loading demo to use a grid and add some 'interactive' demos that show a progress bar that updates and a spinner that shows while the update is in progress


debois:
Thanks—really nice! I took the liberty of merging the interactive and non-interactive bits.

PR #60, opened 2016-06-18
Initial implementation of Footer Component 

### LINKED ISSUE #55: Implement Footer component
See [here](https://getmdl.io/components/index.html#layout-section/footer). CSS only.

vipentti:
Initial implementation of a `Footer` component to resolve #55 

Example use: 
```
import Material.Footer as Footer 

footer : Html m
footer = Footer.mini []
           [ Footer.left [] 
               [Footer.logo [] [Footer.text "Footer Left Section"]]
           , Footer.right []
               [Footer.logo [] [Footer.text "Footer Right Section]]
           ]
```

Open to suggestions and feedback on the both the external API and the internal structure.

The idea was to automate prefixing of the various classes where you have same basically the same element, but with different prefixed class depending on the footers' type. 

e.g. `mdl-mega-footer__left-section`  vs  `mdl-mini-footer__left-section`

This way the user would not have to worry about using the right function, the component would handle it by default. 

The API also exposes `Footer.wrap` function to wrap normal Html elements into footer specific content for cases where you want to have a more customized footer.


src/Material/Footer.elm
```diff
@@ -0,0 +1,342 @@
...
+prefix tp =
+  case tp of
+    Mini -> "mdl-mini-footer"
+    Mega -> "mdl-mega-footer"
+
+separator : Type -> String
+separator tp =
+  case tp of
+    Mini -> "__"
+    Mega -> "__"
```
debois: This looks like you don't need to switch?

src/Material/Footer.elm
```diff
@@ -0,0 +1,339 @@
...
+type FooterProperty = FooterProperty
+
+{-| TODO
+-}
+type alias Property m =
+  Options.Property FooterProperty m
+
+{-| TODO
+-}
+type Content a
```
debois: Do we need this distinction? It is not possible to let everything just be `(Html m)` and do the rest with properties?

debois:
But otherwise, very neat!

If you don't mind, I'd like it if the elm-mdl sources adhere to the [elm style guide](http://elm-lang.org/docs/style-guide), especially the bit about [spacing between top-level declarations](http://elm-lang.org/docs/style-guide#declarations).

src/Material/Footer.elm
vipentti: I think what I had some problems with was trying to apply the prefixes based on the Type of the footer, which is not known until the call to `footer` so the actual `Html m` nodes should not be actually created until that point because currently there is no way to access properties of an existing node.

I was entertaining some thoughts on some sort of `lazy DOM` that would basically wrap existing `Html m` nodes but not create them until explicitly called, this would allow you to insert properties and "modify" them in cases like this.

In this case I think you could do without the `Content m` but the api would would need to be more explicit e.g. `miniLeft` that would apply the specific class

vipentti:
I don't mind at all, I'll apply some formatting

demo/Demo/Footer.elm
```diff
@@ -0,0 +1,229 @@
...
+        ]
+    ]
+
+  , div [Html.style [("margin-top", "60px")]] []
+  , p [] [text "An example of a custom mini footer with custom content"]
+  , div []
+    [ Footer.mini [Options.cs "custom-footer"]
+        [ Footer.wrap <|
+            Html.ul []
+              [ Html.li [Html.class "mdl-mini-footer__social-btn"]
```
debois: See longer comment, point (3).

debois:
Cool! 

A few thoughts about the API.

1. I really like your idea of something lazy-like; I think we should pursue this. If I understand you correctly,  the current `Content` type could probably become something along the lines

    type alias Content = Type -> Html m

This way, you could avoid the slightly award Regex matching to substitue prefixes, and instead simply apply such `Content` functions. 

2. The current API allows me to nest footers. E.g.,:

    nestedFooter = footer [] [ footer [] [...] ]

This is probably not something we want to allow. Looking at the MDL docs, it seems to me that mega-footers always at most one top-, middle- and bottom-section, each of which in turn contain some number of left-, right- and dropdown sections. Would it make sense to have footers take a correspondingly rigid argument, a la 

    type alias MegaFooter = 
      { top : Maybe Section
      , middle : Maybe Section 
      , bottom : Maybe Section 
      }

Then convert some of the existing constructors to this `Section` type. Do you think this would be a sensible idea?

3. In the demo, you add the class `mdl-mini-footer__social-btn` manually. Is it possible we could do that with a property instead? A similar question goes for links and lists. 

4. In the Layout component, I similarly wrap links and other elements. I don't like that `*.link` is proliferating, i.e., `Layout.link`, `Footer.link`, `Menu.link`, but I don't see a good way around it. It's the same situation for `onClick` handlers. Any suggestions?

vipentti:
Hey! 

Thanks for the feedback! Still new to Elm so really appreciate the ideas and the help.


(1). I think something like this, not sure how exactly it would work and look.
I think the idea was that currently there is no way to access the properties and content of a node, so this way you could insert properties/attributes into the an element before actually creating it. 

An alternative could be just making functions that take the dom node function as well as the parameters as its parameter and perform the operations there.

```elm 

doStuff : (List (Attribute msg) -> List (Html msg) -> Html msg) -> (List (Attribute msg)) -> (List (Html msg)) -> Html msg 
doStuff element attributes content = 
    let 
        editedAttributes = update attributes
    in 
        element editedAttributes content

-- In client code

stuff = 
    doStuff div [class "outer-div"] [div [class "inner-div"] [ text "another" ]]

``` 

This does have the problem that it does not work with nested content, so if you wanted to perform some operations on the inner div it would not be possible without another call to the function in the inner content.

---

(2). Yeah, I thought about the rigid arguments for footers but I wasn't sure initially how much freedom we'd like to give the users. I think if we want to make the API more explicit and follow the actual guidelines of the MDL docs then I do think this is better. 
This would make the `footer` API more like the layout API.

I'll try and see how this will work out.

Also this way you could actually use the compiler to help you build your footer the right way.

---

(3). This should be possible, and probably better.



(4). I think this is something would be nice if it could be 'centralized', much like the Html.Events and Html.Attributes, but I'm at a loss currently on how it could be done. 

Maybe eventually something like :
```elm
import Material.Attributes exposing (href) 
import Material.Events exposing (onClick)
```

I'll be sure to keep this one in mind

vipentti:
Did a major refactor of the Footer API.

It is now much more strongly typed and more rigid.

For example: 

```elm 
{-| MiniFooter consists of two sections
-}
type alias MiniFooter a =
  { left : Maybe (Section a)
  , right : Maybe (Section a)
  }


-- Usage

import Material.Footer as Footer

footer : Html Msg
footer =
    Footer.mini []
        { left =
            Footer.left []
                [ Footer.logo [] [ Footer.text "Mini Footer Example" ]
                , Footer.links [] <| createLinks 3
                ]

        , right =
            Footer.right []
                [ Footer.logo [] [ Footer.text "Mini Footer Right Section" ]
                , Footer.socialButton [Options.css "margin-right" "6px"] []
                , Footer.socialButton [Options.css "margin-right" "6px"] []
                , Footer.socialButton [Options.css "margin-right" "0px"] []
                ]
        }
        
``` 

The internals did not change too much yet, still running some specific things with applying prefixes and such.

I think this is a better API for following the MDL guidelines.

Looking forward to more suggestions and feedback.

src/Material/Footer.elm
```diff
@@ -0,0 +1,590 @@
+module Material.Footer
+  exposing
+    ( Type(..)
+    , Property
+    , Content(..)
```
debois: Do we need to expose the constructors? If we don't, we can later change the internal representation (if we like) without inducing a major version bump.

src/Material/Footer.elm
```diff
@@ -0,0 +1,590 @@
...
+  Content
+    { styles = headingClass :: styles
+    , content = content
+    , elem = Html.h1
+    }
+
+
+{-| Wraps `Html.text` element to `Content`
+-}
+text : String -> Content m
```
debois: I don't think we should re-implement parts of the Html.* module whenever we can avoid it. Sticking with `wrap` (which should maybe be called `html`?) would be better.

src/Material/Footer.elm
```diff
@@ -0,0 +1,590 @@
...
+      , content : List (Content a)
+      }
+
+
+{-| Strongly typed `MiddleSection` in a footer
+-}
+type MiddleSection a
+  = MiddleSection
+      { props : List (Property a)
+      , content : List (Dropdown a)
```
debois: Only the middle section can have Dropdown, and the middle section can have no other type of section. Is this true?

debois:
Overall, this seems very nice! I like the additional structure imposed by the type of `Footer` and its constituent components. 

I made a few comments directly in the source; let me know when you're ready for me to merge.

src/Material/Footer.elm
vipentti: I think it might not actually be true, I think I was basing it on the examples where they placed the dropdowns in the middle, which I think is the preferred place for them, but perhaps not something we need to specifically enforce in this case.

src/Material/Footer.elm
vipentti: I don't think we need to expose the constructors.

src/Material/Footer.elm
vipentti: I think `text` is unnecessary here, `heading` I think might be okay, it is very specifically an h1 element with the class `mdl-mega-footer__heading`

> Defines a heading as a mega-footer heading	
> Required on h1 element inside drop-down section

vipentti:
Thanks a lot for the feedback! I think the `Footer` is now in a pretty good place usage wise.

debois:
Very good, thanks!

PR #64, opened 2016-06-22
Initial implementation of Tooltip Component 

### LINKED ISSUE #54: Implement Tooltip component
See [here](https://getmdl.io/components/index.html#tooltips-section).

vipentti:
Initial implementation of a Tooltip component to resolve #54.

Example usage: 

```elm
            
import Material.Tooltip as Tooltip
import Material.Icon as Icon

-- Note the index in both the Render as well as the mouse event handlers

tooltip : Model -> Html Msg
tooltip model =
    div []
    [ Icon.view "add" [ Tooltip.attach Mdl [0] ]
    , Tooltip.render Mdl [0] model.mdl
        []
        [text "Default tooltip"]
    ]
                   
```

Currently there is only one problem that I'm not sure how to best solve: Currently tooltips are always contained in a `div` element which might not be always what you want. However, I'm not sure on the best way to allow users to select the container element for the tooltip.

One option would be  adding a configurable property to Config e.g.

```elm 
{-| Tooltip config
-}
type alias Config a =
  { size : Size
  , position : Position
  , container : HtmlElementFunction a 
  }


-- Set the element
container : HtmlElementFunction a -> Property m
container elem = 
  (\options -> { options | container = elem })



-- Usage 
tooltip : Model -> Html Msg
tooltip model =
    div []
    [ Icon.view "add" [ Tooltip.attach Mdl [0] ]
    , Tooltip.render Mdl [0] model.mdl
        -- Set the container to span element
        [ Tooltip.container Html.span ]
        [text "Default tooltip"]
    ]

```

Open to feedback and suggestions.


src/Material/Tooltip.elm
```diff
@@ -0,0 +1,523 @@
...
+    , Msg(..)
+    , update
+    , view
+    , render
+    , DOMState
+    , left
+    , right
+    , top
+    , bottom
+    , default
```
debois: I don't think we need a `default` property: you specify it by not doing anything else. None of the other components have it.

src/Material/Tooltip.elm
```diff
@@ -0,0 +1,523 @@
...
+    , large
+    , Property
+    , onMouseEnter
```
debois: I don't think we should expose `onMouseEnter` / `onMouseLeave`; user's should always just use `attach`, yes?

src/Material/Tooltip.elm
```diff
@@ -0,0 +1,523 @@
...
+
+    marginTop =
+      -1 * (offsetHeight / 2)
+
+    out =
+      case pos of
+        Left ->
+          { left = props.left - offsetWidth - 10
+          , top =
+              if (top + marginTop < 0) then
```
debois: Very difficult to read. Maybe write the `if` on a single line:

    , top = if (top + marginTop < 0) then 0 else top

Same multiple times below.

src/Material/Tooltip.elm
```diff
@@ -0,0 +1,523 @@
...
+      (toString f) ++ "px"
+
+    pos =
+      if model.isActive then
+        calculatePos config.position model.domState
+      else
+        defaultPos
+  in
+    Options.styled div
+      (cs "mdl-tooltip"
```
debois: No need to manually construct the list with `::`; I think the following is clearer:

    [ cs "mdl-tooltip" 
    , cs "is-active" `when` model.isActive
    , ...

src/Material/Tooltip.elm
```diff
@@ -0,0 +1,523 @@
...
+      Parts.embedUpdate get set1 update
+  in
+    Parts.pack embeddedUpdate
+
+
+{-| Mouse enter event handler for Parts version
+-}
+onMouseEnter : (Parts.Msg (Container b) -> d) -> Parts.Index -> Attribute d
+onMouseEnter lift idx =
+  Html.Events.on "mouseenter" (Json.map (Enter >> ((pack idx) >> lift)) stateDecoder)
```
debois: This is such a clever use of the parts mechanics. I really like it.

src/Material/Tooltip.elm
```diff
@@ -0,0 +1,523 @@
...
+> site's content or function. Their design and use is an important factor in the
+> overall user experience. See the tooltip component's Material Design
+> specifications page for details.
+
+See also the
+[Material Design Specification](https://material.google.com/components/tooltips.html).
+
+Refer to [this site](http://debois.github.io/elm-mdl#/tooltips)
+for a live demo.
+
```
debois: Add docs here indicating that users must _both_ use `attach` to install handlers on the target element, as well as call `render` on the following sibling element.

src/Material/Tooltip.elm
```diff
@@ -0,0 +1,523 @@
...
+  let
+    createPath depth =
+      let
+        parents =
+          List.repeat depth "parentElement"
+      in
+        ([ "target" ] ++ parents ++ [ "nextSibling" ])
+
+    paths =
+      List.map createPath [0..4]
```
debois: You look for the tooltip element among the next 5 siblings, right? Is there are reason we don't just demand that it's the following element?

src/Material/Tooltip.elm
```diff
@@ -0,0 +1,523 @@
...
+          Json.succeed True
+        else
+          Json.succeed False
+      )
+
+
+{-| Decodes a DOMState from a DOM event
+-}
+stateDecoder : Json.Decoder DOMState
+stateDecoder =
```
debois: Neat.

src/Material/Tooltip.elm
```diff
@@ -0,0 +1,523 @@
...
+
+
+{-| Large tooltip
+-}
+large : Property m
+large =
+  Options.set (\options -> { options | size = Large })
+
+
+{-| Maps a Html.Attribute to Material Property
```
debois: Do we need this? I couldn't find a use of it, and I really don't want to give users the ability to install arbitrary attributes on elm-mdl constructs—I've seen them in the past use it in weird ways that break various things, then file bug reports here.

demo/Demo/Tooltip.elm
```diff
@@ -0,0 +1,222 @@
...
+                    import Material.Icon as Icon
+
+                    -- Note the index in both the Render as well as the mouse event handlers
+
+                    tooltip : Model -> Html Msg
+                    tooltip model =
+                      div []
+                        [ Icon.view "add" [ Tooltip.attach Mdl [0] ]
+                        , Tooltip.render Mdl [0] model.mdl
+                            [Tooltip.default]
```
debois: Should just

     []

yes?

demo/Demo/Tooltip.elm
```diff
@@ -0,0 +1,222 @@
...
+-- VIEW
+
+
+demoTooltip : (Html a, String) -> Grid.Cell a
+demoTooltip (tooltip, description) =
+  Grid.cell
+    [Grid.size Grid.All 4]
+    [ div [style [("text-align", "center")]]
+        [tooltip]
+    , code description
```
debois: Description should be 

    , p [] [ text description ] 

(Code-blocks should always be, well, code :)

demo/Demo/Tooltip.elm
```diff
@@ -0,0 +1,222 @@
...
+          [ demoTooltip
+            (div []
+               [ Icon.view "add"
+                   [ Tooltip.attach Mdl [0] ]
+
+               , Tooltip.render Mdl [0] model.mdl
+                   [Tooltip.default]
+                   [text "Default tooltip"]
+               ]
+            , """
```
debois: Generally need more space between the element with the tooltip and the description below. Especially Button.

debois:
Very cool! 

I made a couple of comments on the source which you may or may not want to take into account. Take a look and let me know what you think?

src/Material/Tooltip.elm
vipentti: I think the problem is that if you have for example the following HTML using MDL tooltips

```html 
<p>HTML is related to but different from 
    <span id="xml"><i>XML</i></span>.
    <span>Here is another span</span>  
</p>
<span for="xml" class="mdl-tooltip">Tooltip</span>
```

Then in this case the first nextSibling for the target would be the other span, so we have to travel back the tree to find the actual tooltip element because there is no easy way to tell how nested the tooltips target actually is.

src/Material/Tooltip.elm
vipentti: `onMouseEnter` and `onMouseLeave` are needed if you want to have a tooltip for a regular `Html m` element.

src/Material/Tooltip.elm
vipentti: You are right, I apparently forgot to remove it.

vipentti:
Thanks a lot for the feedback! 

Implemented most of the suggestions and added support for changing the container element in config 

e.g. 

```elm

tooltip = 
  div []
     [ Icon.view "face"
         [ Tooltip.attach Mdl [4] ]     
     , Tooltip.render Mdl [4] model.mdl
       [ Tooltip.large
       , Tooltip.top
       , Tooltip.container Html.span]
         [ img [src "assets/elm.png", width 24, height 24] []
         , text "A target with a tooltip containing both an image and text."
         ]
     ]
```

Issue #61, opened 2016-06-21
Allow Change of Menu Icon


SauceWaffle:
debois,

Needed to have multiple menus on a page so needed to be able to change what icon displays for them. Attached my version of Menu.elm with the changes, basically just imported Icon, setup an option and changed line 347 to set the Icon as whatever was sent in config. 
[Menu.elm.txt](https://github.com/debois/elm-mdl/files/325893/Menu.elm.txt)

Usage: 
"
userMenu : Models.Model -> Html Messages.Msg
userMenu model =
  div []
  [
    Menu.render MDL [0] model.mdl
    [ Menu.bottomRight
    , Menu.ripple
    , Menu.icon "account_circle"
    ]
    [ Menu.Item False True <| div [] [ text "My Profile" ]
    , Menu.Item False True <| div [] [ text "Log In" ]
    ]

  ]
"

debois:
Looks good! Could you submit a PR?

SauceWaffle:
"It seems you do not have permission to push your changes to this
repository" :(

[image: Inline image 1]

debois:
GitHub has a very neat way to do this, but you have to do it in a particular way. See the GitHub docs [here](https://help.github.com/articles/using-pull-requests/).

debois:
Do you have time to upgrade the .txt file to an actual PR?

SauceWaffle:
Very sorry. I live in Cleveland and our team just won the basketball
championship. It's been crazy and I wasn't near a computer all day
yesterday. I'll have the PR to you today.

debois:
Well, congratulations, then :)

I'll take a look at your PR.

debois:
Closed by #65.

PR #65, opened 2016-06-23
Allow Change of Menu Icon (#61)

SauceWaffle:
Added option to set menu icon


debois:
Great, thanks—welcome to elm-mdl :)

PR #66, opened 2016-06-24
Initial implementation of Tabs Component 

### LINKED ISSUE #56: Implement Tabs component
See [here](https://getmdl.io/components/index.html#layout-section/tabs).

vipentti:
Initial implementation of `Tabs` component to resolve #56

Took inspiration from Layout tabs, but with strong opaque types for `tabs`, `labels`  and `content`.

One thing I'm not 100% sure about is having the user provide the `onSelectTab` and `selectTab model.tab` for selecting tab. You could do them in the `Tabs.Model` but then the user would not be able to get messages about when a tab has been selected etc.

Example use: 

```elm
import Material.Tabs as Tabs

tabs : Model -> Html Msg
tabs model =
    Tabs.render Mdl [0] model.mdl
        [ Tabs.ripple
        , Tabs.onSelectTab SelectTab
        , Tabs.selectTab model.tab
        ]
        [ Tabs.tab
            { label = Tabs.label [] [text "Tab One"]
            , content = Tabs.content [] [text "Tab One content"]
            }

        , Tabs.tab
            { label = Tabs.label [] [text "Tab Two"]
            , content = Tabs.content [] [text "Tab Two content"]
            }
        ]
```

Open to suggestions and feedback.

### DIFF
--- a/src/Material/Tabs.elm
+++ b/src/Material/Tabs.elm
@@ -1,0 +1,238 @@
+module Material.Tabs exposing
+  (..)
+
+-- TEMPLATE. Copy this to a file for your component, then update.
+
+{-| From the [Material Design Lite documentation](http://www.getmdl.io/components/#TEMPLATE-section):
+
+> ...
+
+See also the
+[Material Design Specification]([https://www.google.com/design/spec/components/TEMPLATE.html).
+
+Refer to [this site](http://debois.github.io/elm-mdl#/template)
+for a live demo.
+
+@docs Model, model, Msg, update
+@docs view
+
+# Component support
+
+@docs Container, Observer, Instance, instance, fwdTemplate
+-}
+
+
+import Platform.Cmd exposing (Cmd, none)
+import Html exposing (Html)
+import Parts exposing (Indexed)
+import Material.Options as Options exposing (cs, when)
+import Material.Helpers as Helpers
+import Material.Ripple as Ripple
+import Html.App
+import Html.Attributes as Html exposing (class)
+import Html.Events as Html
+
+import Dict exposing (Dict)
+
+-- MODEL
+
+
+{-| Component model.
+-}
+type alias Model =
+  { ripple : Ripple.Model
+  , activeTab : Int
+  }
+
+
+{-| Default component model constructor.
+-}
+defaultModel : Model
+defaultModel =
+  { ripple = Ripple.model
+  , activeTab = 0
+  }
+
+
+-- ACTION, UPDATE
+
+
+{-| Component action.
+-}
+type Msg
+  = Select Int
+  | Ripple Ripple.Msg
+
+
+{-| Component update.
+-}
+update : Msg -> Model -> (Model, Cmd Msg)
+update action model =
+  case action of
+    Select idx ->
+      ({ model | activeTab = idx }, none)
+    Ripple action' ->
+      let
+        (nextRipple, cmd) = Ripple.update action' model.ripple
+      in
+        ({ model | ripple = nextRipple }, Cmd.map (Ripple) cmd)
+
+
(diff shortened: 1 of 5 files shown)


debois:
Very good!

I'm not sure it's a good idea having the user provide contents for all tabs and not just the selected one:
1. The contents of every tab will be re-computed on every call to `view`. This is potentially expensive in CPU and on mobile battery.
2. Computing the contents of non-active tabs might require HTTP access etc. (dynamically loaded contents)

Have you thought about how your current implementation interacts with routing? I'd expect it to be very common that URLs affect the selected tab. 

I like the `onSelectTab` /  `selectedTab` approach—this is the principle behind all elm-mdl APIs: You specify as much as possible as Options; in particular, you register handlers to get notified when stuff happens. If at all possible, we want the user not to need to care about the Model of a component.

src/Material/Tabs.elm
```diff
@@ -0,0 +1,333 @@
...
+
+
+
+-- ACTION, UPDATE
+
+
+{-| Component action.
+-}
+type Msg
+  = NoOp
```
debois: `NoOp` is not used anyplace, yes? Maybe get rid of it.

debois:
Very good!

I'm not sure it's a good idea having the user provide contents for all tabs and not just the selected one:
1. The contents of every tab will be re-computed on every call to `view`. This is potentially expensive in CPU and on mobile battery.
2. Computing the contents of non-active tabs might require HTTP access etc. (dynamically loaded contents)

Have you thought about how your current implementation interacts with routing? I'd expect it to be very common that URLs affect the selected tab. 

I like the `onSelectTab` /  `selectedTab` approach—this is the principle behind all elm-mdl APIs: You specify as much as possible as Options; in particular, you register handlers to get notified when stuff happens. If at all possible, we want the user not to need to care about the Model of a component.

vipentti:
Hey! 

Thanks for the feedback again!

(1). I think you are right. It just means the user is responsible setting the content based on the selected tab, which should not require too much since they already have to provide `onSelectTab` and `selectTab` options.

(2). I'll try and simplify the API to go along the (1) changes.  

(3). Regarding routing: I did not actually consider that, I would assume the user would have to take care of managing it using `selectTab`. I think MDL somewhat separates navigation based `Tabs` from `Tabs` component. The tabs used for navigation are in Layout, and the tabs used to display different content in the same screen space are part of the `Tabs` component.

vipentti:
Hey!

Implemented a new `Tabs` API based on your suggestions. I think it is simpler to use and less verbose.


**(1).** `textLabel` is a simple utility function to create a label with just text, as this is probably the most common case for tab labels. 

**(2).** I decided to keep `Label m` opaque type. I like the idea of using elm compiler help us enforce certain things that are not possible to enforce in using regular MDL CSS.   

**(3).** One thing I'm still not 100% on is the last argument to the render function, the content of the active tab. I do like the consistency of providing `List (Html m)` but at the same time maybe just providing a single `Html m` element would be enough. The provided `content` will be wrapped in a `div` element internally with the appropriate classes.

Example use:

```elm
     import Material.Tabs as Tabs

     tabs : Model -> Html Msg
     tabs model =
       Tabs.render Mdl [0] model.mdl
           [ Tabs.ripple
           , Tabs.onSelectTab SelectTab
           , Tabs.selectTab model.tab
           ]
           [ Tabs.label [] [text "Tab One"]
           , Tabs.textLabel [] "Tab Two"
           ]
           [ case model.tab of
               0 -> div [] [text "Content of tab one"]
               1 -> div [] [text "Content of tab two"]
               _ -> div [] []
           ]
```

Open to more feedback and suggestions as always!

aforemny:
Thank you very much! :)

PR #72, opened 2016-07-07
Initial implementation of Slider Component 

### LINKED ISSUE #69: Implement Slider component
See [here](https://getmdl.io/components/index.html#sliders-section).  Some javascript [involved](https://github.com/google/material-design-lite/blob/master/src/slider/slider.js).

vipentti:
Initial implementation of a `Slider` component to resolve #69.

Did a major refactor of the API based on the [discussion](https://github.com/debois/elm-mdl/issues/69).

The component does not currently use or support `Parts` component model because the component itself has no need for `update` because of the user having to provide both `onChange` and `value` calls to update the slider.

`Slider` does not currently work properly on [Microsoft Edge](https://github.com/google/material-design-lite/issues/1625). It also does not currently look the best on regular Internet Explorer due to flexbox. However, it is functional.

Example use: 

```elm 
import Material.Slider as Slider

slider : Model -> Html Msg
slider model =
    p [ style [ ("width", "300px") ] ]
    [ Slider.slider
        [ Slider.onChange SliderMsg
        , Slider.value model.value
        ]
    ]
```

This follows the `MDL` of wrapping the actual slider in a separate element that has a width specified.

### DIFF
--- a/demo/Demo.elm
+++ b/demo/Demo.elm
@@ -33,6 +33,7 @@ import Demo.Layout
 import Demo.Footer
 import Demo.Tooltip
 import Demo.Tabs
+import Demo.Slider
 --import Demo.Template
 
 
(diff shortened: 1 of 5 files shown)


debois:
Very good—I merged this, leaving open issues #74 and #75 about IE/Edge compatibility.

PR #85, opened 2016-07-17
Remove bad "]" in code example

slegrand45:


debois:
Very good! Keep 'em coming :)

PR #86, opened 2016-07-17
Lists

hakonrossebo:
This is my initial implementation of Lists. I've created demo lists that seems to replicate the mdl examples ok. I have named the component Lists instead of List to avoid naming conflicts with core list. Please suggest any improvements.


debois:
Hi Håkon, very nice!

I'll make some minor comments directly in the source. See if they make sense to you.

src/Material/Lists.elm
```diff
@@ -0,0 +1,194 @@
...
+    Options.set <| \self -> { self | listItemLinesVariant = lines }
+
+
+{-| List item
+-}
+li :
+    List (Property ListItemConfig m)
+    -> List (Html m)
+    -> Html m
+li options nodes =
```
debois: Maybe make a constructor for each of the three list variants (item, item--two-line, item--three-line)? It would make it easier for user's and it'd be really easy to implement: 

    li options = 
      Options.styled li (cs "mdl-list__item" ++ options)

    li2 options = 
      Options.styled li (cs "mdl-list__item--two-line" ++ options)

Then you can also get rid of the `ListItemLinesVariant`

src/Material/Lists.elm
```diff
@@ -0,0 +1,194 @@
...
+                    cs "mdl-list__item--three-line"
+            , cs "mdl-list__item"
+            ]
+            []
+            nodes
+
+
+{-| Defines the primary content sub-division
+-}
+itemPrimaryContent : Property a m
```
debois: Maybe lose the `item` prefix. The user is going to write `List.primaryContent` anyway, which is probably definite enough. Similar comment for all the following values.

src/Material/Lists.elm
```diff
@@ -0,0 +1,194 @@
...
+> Lists present multiple line items vertically as a single continuous element.
+> Refer the Material Design Spec to know more about the content options.
+
+See also the
+[Material Design Specification]([https://material.google.com/components/lists.html).
+
+Refer to [this site](http://debois.github.io/elm-mdl#/lists)
+for a live demo.
+
+Example use:
```
debois: Either give a proper example, or just refer to the demo. I'm kind of prefer the the demo having a good example including code.

demo/Demo/Lists.elm
```diff
@@ -0,0 +1,307 @@
...
+            ]
+        ]
+
+
+twoLines : Model -> Html Msg
+twoLines model =
+    Lists.ul [ Options.css "width" "300px" ]
+        [ Lists.li [ Lists.twoLine ]
+            [ Options.span [ Lists.itemPrimaryContent ]
+                [ Options.styled Html.i [ Lists.itemAvatar ] [ Icon.view "person" [ Icon.size36 ] ]
```
debois: Doesn't this produce nested `i` tags (one we made here, one from `Icon.view`)? Couldn't we just do

   [ Icon.view "person" [ Icon.size36, Lists.itemAvatar ] 

?

demo/Demo/Lists.elm
```diff
@@ -0,0 +1,307 @@
...
+                , Options.span [] [ text "Bryan Cranston" ]
+                , Options.span [ Lists.itemSubTitle ] [ text "62 Episodes" ]
+                ]
+            , Options.span [ Lists.itemSecondaryContent ]
+                [ Options.span [ Lists.itemSecondaryInfo ] [ text "Actor" ]
+                , Options.styled Html.a [ Lists.itemSecondaryAction ] [ Icon.view "star" [] ]
+                ]
+            ]
+        , Lists.li [ Lists.twoLine ]
+            [ Options.span [ Lists.itemPrimaryContent ]
```
debois: Are `itemSecondaryContent` and others always `span`, or could they conceivably sometimes be sometimes else, like `div`? If they are always `span`, maybe define them as such:

    itemSecondaryContent options = 
      Options.span (cs "mdl-list--secondary-content" ++ options)

demo/Demo/Lists.elm
```diff
@@ -0,0 +1,307 @@
...
+    Lists.ul [ Options.css "width" "650px" ]
+        [ Lists.li [ Lists.threeLine ]
+            [ Options.span [ Lists.itemPrimaryContent ]
+                [ Options.styled Html.i [ Lists.itemAvatar ] [ Icon.view "person" [ Icon.size36 ] ]
+                , Options.span [] [ text "Bryan Cranston" ]
+                , Options.span [ Lists.itemTextBody ] [ text "Bryan Cranston played the role of Walter in Breaking Bad. He is also known for playing Hal in Malcom in the Middle." ]
+                ]
+            , Options.span [ Lists.itemSecondaryContent ]
+                [ Options.span [ Lists.itemSecondaryInfo ] [ text "Actor" ]
+                , Options.styled Html.a [ Lists.itemSecondaryAction ] [ Icon.view "star" [] ]
```
debois: SImilarly, are `...Action` always `a`?

src/Material/Lists.elm
```diff
@@ -0,0 +1,194 @@
...
+import Html.Events as Html
+import Html exposing (Html, Attribute)
+import Parts exposing (Indexed)
+import Material.Options as Options exposing (Property, Style, cs, nop)
+
+
+{-| Main list function
+-}
+ul :
+    List (Property {} m)
```
debois: Why `Property {} m` and not `Property a m`? Do we need to restrict?

src/Material/Lists.elm
hakonrossebo: Separate constructors seems to be better. I tried to use your example code, but the compiler complained when I tried different signatures. Had to use Options.apply for now. Could you help me get this right?

src/Material/Lists.elm
hakonrossebo: Done

src/Material/Lists.elm
hakonrossebo: Done. Refers to demo

demo/Demo/Lists.elm
hakonrossebo: Had an issue with nested i tags. Thanks.

demo/Demo/Lists.elm
hakonrossebo: I think this is flexible. We could  of course provide a default implementation.

demo/Demo/Lists.elm
hakonrossebo: I think this is flexible. We could  of course provide a default implementation.

src/Material/Lists.elm
hakonrossebo: I tried with a m, but did not manage to compile. Not sure how to change this correctly.

src/Material/Lists.elm
```diff
@@ -0,0 +1,174 @@
...
+    -> List (Html m)
+    -> Html m
+ul options nodes =
```
debois: Simpler implementation of the same thing:

   ul options = 
      Options.styled Html.ul (cs "mdl-list ++ options)

src/Material/Lists.elm
```diff
@@ -0,0 +1,174 @@
...
+
+
+
+{-| Default List item
+-}
+li :
+    List (Property {} m)
+    -> List (Html m)
+    -> Html m
+li options nodes =
```
debois: Similar to ul.

src/Material/Lists.elm
```diff
@@ -0,0 +1,174 @@
...
+        Options.apply summary
+            Html.li
+            [ cs "mdl-list__item" ]
+            []
+            nodes
+
+{-| List item with two lines
+-}
+li2 :
+    List (Property {} m)
```
debois: Similar to ul.

src/Material/Lists.elm
```diff
@@ -0,0 +1,174 @@
...
+            nodes
+
+{-| List item with three lines
+-}
+li3 :
+    List (Property {} m)
+    -> List (Html m)
+    -> Html m
+li3 options nodes =
+    let
```
debois: Similar to ul

hakonrossebo:
Updated with changes from latest comments. I had to update and commit Page.html in demo since I had used a wrong elm-make command last time and performed a commit.

debois:
Very good, thanks. I'll take a look at the property types. I have been thinking about the API, but I need to experiment a bit. Would you mind if I take it from here, or would you prefer to keep working on this yourself based on discussions here (both are fine for me)?

hakonrossebo:
Please take over from here - I'm sure it's more efficient when experimenting :)

debois:
Excellent! I've spiffed up the demo a bit, and made some minor adjustments to the API—the final demo is [here](https://debois.github.io/elm-mdl/wip.html#lists)

hakonrossebo:
Great. I like the demo enhancements and api changes :)

PR #89, opened 2016-07-19
Improve example documentation

jvoigtlaender:
* The `Platform` import is superfluous.
* The Msg constructor was given in wrong capitalization.

Still, though: The item
> An initial model (`(Button.model True)`---a button with a ripple animation.

is confusing, since nothing like `(Button.model True)` acually occurs in the file.

### DIFF
--- a/examples/Component.elm
+++ b/examples/Component.elm
@@ -1,7 +1,6 @@
 import Html.App as App
 import Html exposing (..)
 import Html.Attributes exposing (href, class, style)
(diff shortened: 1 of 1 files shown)


debois:
Thanks again. I updated the outdated comments in examples. (Depending on what you were looking for when you came across the old examples, you might want to take a second look.)

jvoigtlaender:
Cool, thanks!

PR #90, opened 2016-07-19
Fix indenting in code example

jvoigtlaender:
... so that it will be properly highlighted on the package site.


debois:
Much appreciated, thanks!

PR #91, opened 2016-07-19
Typography

### LINKED ISSUE #18: Typography module
Material design defines [typography styles](https://www.google.com/design/spec/style/typography.html#typography-styles) in great detail. MDL apparently (but undocumentedly) [implements this](https://github.com/google/material-design-lite/blob/master/src/typography/_typography.scss#L123-L240). 

We should have the corresponding Styles in a Typography module with demo page etc.

vipentti:
Initial implementation of `Typography` module to resolve #18.

Right now it simply returns the class strings, that the user must manually apply to with a `Html.class`. 

Example use: 

```elm 
import Material.Typography as Typo
import Html exposing (p)
import Html.Attributes exposing (class)

example : Html msg 
example = p [ class Typo.display4 ] 
            [text "Light 112px"]
```

There is also an utility to combine many classes

```elm
import Material.Typography as Typo
import Html exposing (p)
example : Html msg 
example = p [ Typo.many [Typo.left, Typo.titleColorContrast] ] 
            [ text "Left align" ]
```  

I wish `Html.class` would work like `Material.Options.cs` allowing you to have multiple `class` declarations in your properties. This would allow for simpler API.  

Open to feedback and suggestions.

### DIFF
--- a/demo/Demo.elm
+++ b/demo/Demo.elm
@@ -34,6 +34,7 @@ import Demo.Footer
 import Demo.Tooltip
 import Demo.Tabs
 import Demo.Slider
+import Demo.Typography
 --import Demo.Template
 
 
@@ -56,6 +57,7 @@ type alias Model =
   , tooltip : Demo.Tooltip.Model
   , tabs : Demo.Tabs.Model
   , slider : Demo.Slider.Model
+  , typography : Demo.Typography.Model
   --, template : Demo.Template.Model
   , selectedTab : Int
   , transparentHeader : Bool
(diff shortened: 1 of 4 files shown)


debois:
Awesome, thanks.

I think it's be better if this used `cs ...` from the Options module; most other Material modules use this approach already. Maybe take a look at how the Elevation module does it?

vipentti:
I thought about doing `cs` for these but I'm not sure what's the best approach to then using  them on normal `Html` elements such as `p`, `span` etc.

I guess we could add `Options.p` like `Options.div` and `Options.span` and then just tell the users to use our functions but I'm not sure if that is the best approach.

vipentti:
Updated, thanks for the feedback!

debois:
Very good, thanks!

Issue #87, opened 2016-07-18
Additional textfield attributes?


kittykatattack:
Hello,

Is it possible to set standard elm-html attributes on MDL textfields, such as `onEnter`, `onBlur`, `onFocus`, `maxlength` and `autofocus`?

debois:
Not presently, but it should be.

We removed the ability to install arbitrary `Html.Attributes` on elm-mdl elements because it sometimes messes up elm-mdl events, and because users kept finding ways to get in trouble with it.

The right way to go would be to implement the handlers needed like the current [`onInput` handler](https://github.com/debois/elm-mdl/blob/master/src/Material/Textfield.elm#L144,L150). Note that `onBlur` and `onFocus` would need to be implemented by `onfocusout/onfocusin` in the div containing the input.

I would look at this, but I have to prioritize Firefox breaking everything all the time. 

Would you care to submit a PR? I'd help out with any questions you may have, either here or on the elm-slack (@debois). If you spend five minutes looking at the [Textfield implementation](https://github.com/debois/elm-mdl/blob/master/src/Material/Textfield.elm), I'm sure you'll find it's something you can hack :)

kittykatattack:
Thank you, Sir!
Unfortunately I do not yet have the skill to submit a PR, but would certainly love to help out in future as my Elm skills improve.
No worries, I can work around this for now :smile: 
Good luck with your work on Firefox (I've also had a lot problems with it recently in another JS library I  maintain.)

debois:
Ok!

In case you change your mind (or suddenly have time on your hands, or someone else wants to have a go):

To add `maxlength` or `autofocus`, you need just mimic the `value` attribute, which is supported. See [here](https://github.com/debois/elm-mdl/blob/master/src/Material/Textfield.elm#L128-L133) and [here](https://github.com/debois/elm-mdl/blob/master/src/Material/Textfield.elm#L307-L309).

To add `on*`, see the comments above.

debois:
Ville did this. It'll be in the next release.

PR #92, opened 2016-07-20
Textfield extra attributes

vipentti:
Had some time on my hands so added some attributes that were discussed in #87.

 ```elm
  , Textfield.render MDL [6] model.mdl
      [ Textfield.label "Default multiline textfield (50 char limit)"
      , Textfield.textarea
      , Textfield.autofocus -- New 
      , Textfield.maxlength 50 -- New
      , Textfield.onBlur (Blur [6]) -- New
      , Textfield.onFocus (Focus [6]) -- New
      ]

```

### DIFF
--- a/demo/Demo/Textfields.elm
+++ b/demo/Demo/Textfields.elm
@@ -10,7 +10,10 @@ import Material.Options as Options
 import Material
 
 import Demo.Page as Page
+import String
 
+import Material.Slider as Slider
+import Material.Typography as Typo
 
 -- MODEL
 
@@ -20,6 +23,8 @@ type alias Model =
   , str0 : String
   , str3 : String
   , str4 : String
+  , str6 : String
+  , length : Float 
   }
 
 
@@ -29,6 +34,8 @@ model =
   , str0 = ""
   , str3 = ""
   , str4 = ""
+  , str6 = ""
+  , length = 50
   }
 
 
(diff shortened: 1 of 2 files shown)


debois:
Very good thanks. Feel like updating the Demo to somehow showcase onBlur/onFocus, maybe add a slider to set maxlength, and maybe add code samples to everything? (It's ok to say no, in that case, I'll just merge this.)

vipentti:
I'll see what I can do to update the demos a bit.

There should be samples for all options except for table-striped and no-wrap I believe.

vipentti:
Added a better maxlength demo with a slider to increase and decrease the maxlength. 

Also removed some incorrect documentation..

Unfortunately, could not come up with a demo for onFocus and onBlur

debois:
Cool, thanks.

debois:
This PR gave us custom handlers on Textfield, but it turned out to be at the price of ripples on Firefox. (We love you, Firefox.) I think that's a bad tradeoff for the v7 branch. Can we revert it?

vipentti:
I assume you mean focus and blur on firefox ?

debois:
Sorry, yes. To be precise: Without a major version bump, we have to chose at most one of: 

- Fix #166 (textfield doesn't react to focus/blur on Firefox)
- Fix #87 (more user-exposed event handlers for textfield)

I think #166 is way more important, so I'd like to revert this PR; or at least somehow fix #166 at the cost of losing functionality of #87 on Firefox.

We'll get back #87 eventually in master (and so v8) when we merge #179 (custom event handlers).

vipentti:
@debois I think I may have a fix for #166  that will break #87 on Firefox (onBlur and onFocus)

PR #109, opened 2016-07-23
Card component

robvandenbogaard:
This is my proposal for the Card component. I've included a demo page. There seems to be a styling issue with subtitles, however I can replicate the odd styling by inserting a subtitle using the inspector on https://getmdl.io/components/index.html#cards-section.


debois:
Excellent, thanks! Welcome to elm-mdl!

Issue #114, opened 2016-07-25
Button Ripple does not vanish in IE


OvermindDL1:
Testing a site in IE 11 (works fine in Chrome, Firefox, and Edge), and the buttons ripple does not vanish, any button.  It does the effect when clicked but then the slightly grayed out ripple stays instead of vanishing, sometimes the ripple does not even go all the way across the button and instead just stops there, sitting.  Checking the CSS in the DOM on the page shows that the "is-visible" class is still on the ripple, yet is not on the buttons that are not yet touched, but once touched they get "is_visible" added and never removed.  If I remove it manually via the IE11 javascript console then they look correct again until the next time it is clicked, where the ripple will animate but then not vanish.

OvermindDL1:
Testing via https://debois.github.io/elm-mdl/ in IE 11, same issue, the ripple div is not set invisible once the animation is complete.

However, it does work properly at https://getmdl.io/components/index.html#buttons-section thus something is missing in this port?

debois:
Just to be sure: This is on 6.0.4?

OvermindDL1:
Correct, from my elm-stuff/exact-dependencies.json:
```json
    "debois/elm-mdl": "6.0.4",
```

debois:
Fixed in 6.0.5.

PR #116, opened 2016-07-26
Temporary fix for #114

vipentti:
IE11 Does not support creating Synthetic Events using `new Event`.

Temporary workaround using deprecated `document.createEvent`.

Tested on newest Firefox, Chrome, IE11 and Edge.

### DIFF
--- a/src/Material/Button.elm
+++ b/src/Material/Button.elm
@@ -200,7 +200,8 @@ blurAndForward : String -> Attribute m
 blurAndForward event = 
   Html.Attributes.attribute 
     ("on" ++ event) 
-    "this.blur(); this.lastChild.dispatchEvent(new Event('touchcancel'));"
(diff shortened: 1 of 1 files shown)


OvermindDL1:
Maybe it would be better to see if 'new Event' works and if so use it, else fallback to the old method?  Or test if the old method still exists and use it, else use new Event?  If the former is possible it would probably be better.

vipentti:
That would be ideal but the inline javascript is getting out of hand already. If there was a better way to create methods in Javascript to be used with Elm without requiring external code then it might be worth it. An alternative is also a [polyfill](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#Polyfill) that might work, but that also requires external code.

`new Event` is the modern way and should be forward compatible and eventually the `createEvent` might be removed so hopefully before that the Elm focus/blur changes are added.

Issue #52, opened 2016-06-15
Implement Dialog component


debois:
See [here](https://getmdl.io/components/index.html#dialog-section).

aforemny:
I took a quick look at the Material Design Lite implementation of dialogs. According to the (dialog section)[https://getmdl.io/components/index.html#dialog-section], dialogs are _not_ supported by the majority of browsers and thus have to be (polyfilled)[https://github.com/GoogleChrome/dialog-polyfill/blob/master/dialog-polyfill.js].

As far as the polyfill is concerned, it seems to deal with at least the following aspects:

1. positioning of the dialog
2. adding a backdrop that cancels mouse events outside the dialog
3. stacking open dialogs
4. setting focus

Apart from that, dialogs are purely cosmetic and trivially re-implemented in Elm.

I propose the following:

In a *first implementation*, we should come up with a single dialog component that satisfies 1. and 2. This would allow people to use simple (singleton) dialogs, and if necessary implement something more sophisticated in their application themselves.

In a *second implementation*, a simple dialog manager could solve 3. by computing z-indices for a stack of dialogs and the backdrop.

As far as 4. is concerned, we have not yet come up with a good way to deal with focus at all. So this should not be a concern for now. One could, however, investigate if the `autofocus` attribute could do anything sensible here.

debois:
I like your plan. 

Would we have really easy path to fully functioning Dialogs if we required the user to load a polyfill, though?

puhrez:
@debois I'm really interested in working on this feature, with the idea of limitted browser support at first then polyfill. 
I'm new to this project and new(ish) to elm, any advice where I could get started?

debois:
Gentlemen, 

@puhrez provided a very nice PR, neatly settling the API on the elm side, taking the first steps of @aforemny's plan, stopping short of adding open/close capabilities. Many thanks, and welcome to elm-mdl!

I've added those capabilities. As it happens, I couldn't find a way to open a modal dialog without calling `dialog.showModal` from Javascript, using the usual inline-JS-in-event-handler hack. I tried to be real careful about everything being defined to prevent our use of JS from blowing up the elm runtime. @vipentti, could you take a look? [These](https://github.com/debois/elm-mdl/blob/master/src/Material/Dialog.elm#L99-L163) are the operative lines. 

The current implementation has the following limitations (I also wrote these in the docs of `Dialog.view`):

- Using  [the
[...]

PR #120, opened 2016-07-27
Add Dialog Component

puhrez:
Address #52 

- [x] Module written
- [ ] Demo written
- [x] Docs written
- [ ] Enforce "mdl-dialog__actions--full-width" only for Actions


puhrez:
Having problem viewing the

debois:
Awesome, thanks!

debois:
I don't think we need to enforce `fullWidth` only for actions, for a couple reasons: 

1. It's not dangerous. You add it the wrong place, worst thing that happens is you get slightly off layout. It might even work. 
2. It's difficult to enforce without going into very restrictive APIs. 
3. There's plenty of other examples in the library where we are lax with options. 

I did update the docs to make clear it belongs on action.

debois:
I added a new issue #126 for improving the demo. This ticks off your last box :)

PR #151, opened 2016-08-05
Add support for custom event handlers on Textfield

### LINKED ISSUE #130: Textfield should support more DOM event handlers
From @OvermindDL1 at https://groups.google.com/d/msg/elm-discuss/pNPVxrz_jlE/z_zaaattAwAJ : 

> Do not see an example or demonstration code that shows this, but how can I get an Enter keypress from a single line `Textfield`, and how can I distinguish it from a Shift+Enter keypress in a multiline `Textfield`, or how can I get the event data?

vipentti:
Resolves #130  

This also refactors the way event handlers are used internally.
Instead of adding the internal handlers on the input element.
The handlers are now added to the containing div using events that
are known to bubble (`input`, `focusin`, `focusout`)

See the demo code for an example for `keyup` event

The user event handlers are now added to the input element.

Open to feedback and suggestions on this one.

### DIFF
--- a/src/Material/Textfield.elm
+++ b/src/Material/Textfield.elm
@@ -78,7 +78,6 @@ import Platform.Cmd
 import Parts exposing (Indexed)
 
 import Material.Options as Options exposing (cs, css, nop, Style)
-import Dict exposing (Dict)
 
 
 -- OPTIONS
@@ -102,7 +101,7 @@ type alias Config m =
   , autofocus : Bool
   , maxlength : Maybe Int
   , style : List (Options.Style m)
-  , listeners : Dict String (Html.Attribute m)
+  , listeners : List (Html.Attribute m)
   }
 
 
@@ -120,7 +119,7 @@ defaultConfig =
   , autofocus = False
   , maxlength = Nothing
   , style = []
-  , listeners = Dict.empty
+  , listeners = []
   }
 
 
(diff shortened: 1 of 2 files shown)


src/Material/Textfield.elm
```diff
@@ -105,6 +111,7 @@ defaultConfig =
   , maxlength = Nothing
   , onBlur = Nothing
```
debois: We no longer need `onBlur`, `onFocus` in the model it seems. Maybe remove?

src/Material/Textfield.elm
```diff
@@ -105,6 +111,7 @@ defaultConfig =
...
   , onFocus = Nothing
+  , listeners = Dict.empty
```
debois: Shouldn't `listeners` be just a `List`?

src/Material/Textfield.elm
```diff
@@ -319,6 +329,24 @@ view lift model options =
...
+      Dict.values config.listeners
+
+    textValue =
+      case config.value of
+        Just str ->
+          [ Html.Attributes.value str ]
+        Nothing ->
+          []
+
+    defaultInput =
```
debois: I think you do this to make sure the field responds to user input even if the parent is not handling inputs. It's a neat idea, but I don't think it'll work, though: Input sets "is-dirty" which is somehow important. (Floating labels, maybe?)

debois:
Very nice :)

I made a few line comments, and I'd like to suggest you make a "yellow-background" part of the demo where you give an example of hooking up a custom handler , which simply reports the selected text. (You'll probably need a an "onmouseup" handler to get mouse selections also.)

Feel free to merge it (preferably into branch v7) when you're happy with it :)

vipentti:
Hey, 

Thanks for the feedback! Updating the PR and hopefully merge a bit later

PR #153, opened 2016-08-05
Extra url fixes

OvermindDL1:
Apparently it was not just the one that I fixed, but *all* of them were broken because the template was wrong.  Fixed them all, fixed the template, tested every-single-one, think I got them all.

Also, there is a lot of inconsistency in the Demo naming.  Some demo's are the same name as the type like Snackbar, Dialog, Grid, and more even though they show multiples, and others are like Footer/Footers, Card/Cards, etc...  Should probably be unified, I'd say to the same name as the type to make lookup and links easier.

For now, url fixes here, template too.

### DIFF
--- a/src/Material/Button.elm
+++ b/src/Material/Button.elm
@@ -30,7 +30,7 @@ See also the
 [Material Design Specification]([https://www.google.com/design/spec/components/buttons.html).
 
 Refer to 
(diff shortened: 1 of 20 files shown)


debois:
Awesome, thanks!

PR #173, opened 2016-08-13
Fix is-focused not being triggered on Firefox

### LINKED ISSUE #166: Textfield has no class 'is-focused' in Firefox
textfield-focus-firefox

vipentti:
Unfortunately, Firefox does not support `focusin` and `focusout` events. 
This works around the issue by adding internal `focus` and `blur`
handlers on the `input` element and then dispatching user provided
`focus` and `blur` handlers in `update`.

Resolves #166

### DIFF
--- a/src/Material/Textfield.elm
+++ b/src/Material/Textfield.elm
@@ -78,6 +78,8 @@ import Platform.Cmd
 import Parts exposing (Indexed)
 
 import Material.Options as Options exposing (cs, css, nop, Style)
+import Material.Helpers as Helpers
+import Dict exposing (Dict)
 
 
 -- OPTIONS
@@ -100,7 +102,7 @@ type alias Config m =
   , autofocus : Bool
   , maxlength : Maybe Int
   , style : List (Options.Style m)
-  , listeners : List (Html.Attribute m)
+  , listeners : Dict String (Decoder.Decoder m)
   }
 
 
@@ -117,7 +119,7 @@ defaultConfig =
   , autofocus = False
   , maxlength = Nothing
   , style = []
-  , listeners = []
+  , listeners = Dict.empty
   }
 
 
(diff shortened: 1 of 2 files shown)


src/Material/Textfield.elm
```diff
@@ -282,25 +285,39 @@ defaultModel =
...
-      { model | isFocused = True }
+    Blur msg ->
+      let
+        model' =
+          { model | isFocused = False }
+        cmds =
+          List.filterMap identity
+            [ msg |> Maybe.map Helpers.cmd ]
+      in
+        (model', Cmd.batch cmds)
```
debois: Maybe lose `cmds` and just do 
```elm
     (model' , msg |> Maybe.map Helpers.cmd |> Maybe.withDefault Cmd.none)
```

src/Material/Textfield.elm
```diff
@@ -349,8 +366,20 @@ view lift model options =
         Nothing -> []
 
 
+    -- These events require special handling because they do not bubble
+    -- Meaning our event handlers for these events have to be on the input-element as well
+    specialEvents =
+      ["blur", "focus"]
+
+    (special, rest) =
+      Dict.partition
```
debois: I think it'd be nicer if you split out focus and blur already in config, then let the `on` property either put handler in general listeners or special places for blur/focus. This avoids the slightly heavy `Dict.partition`. 

(You could also argue that this is totally irrelevant and leave it as is.)
(I'm starting to be a bit concerned about overhead of `Options` in general; maybe I'm overreacting.)

src/Material/Textfield.elm
```diff
@@ -366,6 +395,21 @@ view lift model options =
         Nothing ->
           Just <| Html.Events.on "input" (Decoder.map (Input >> lift) targetValue)
 
+
+    specialDecoder evtName =
+      case Dict.get evtName special of
```
debois: "Later lookup"

debois:
I like this. I made a few line comments, in particular, I think `listeners` in `Config` should be a list and handlers for `blur` / `focus` separated out—you might avoid a little bit of overhead. But that's neither here nor there. Merge it when you're happy with it :)

debois:
Superceded by #179.

Issue #164, opened 2016-08-10
Custom event handlers for components


vipentti:
Hello there!

We had some discussion about custom event handlers in #130 and we added support for `Textfield.on` in 7.3.0. 

I'd like to open a discussion about possibly adding more general support for `on` and perhaps even `onWithOptions` for MDL components.

Enabling users to add custom event handlers does have a fair chance of breaking things with certain events such as `blur` and `focus` so this does need to be taken into consideration when it comes to enabling user defined events.

Another challenge is where to actually put the user provided event handlers in components. Some components are wrapped in `div` elements such as Textfield which allow us to put our internal event handlers on the container and the user provided ones on the internal `input` or `textarea`, and use the bubbling of certain events to our advantage. However, this is not the case with all the components. 

If there are any particular use cases for custom event handlers that are not currently possible it would be awesome to hear about them.

debois:
If we could do this, it'd be great. I think the approach you took in #151 could reasonably be applied to other components. I think we need to understand and fix #166 before we proceed though. 

We might want to swap the approach of #151 and add user handlers to containers and our handlers directly on elements, that'll remove the problem of user handlers interfering. This'll generally mean that user's won't get keyboard events (right?). I'm unsure what's the right tradeoff no-keyboard, no-interference vs. keyboard, possible-interference.

I think the way to proceed would be to:
1. Fix #166.
2. Make a list of components that would benefit from arbitrary DOM handlers.
3. Go through the list, figure out which components have containers we can put user handlers on. 
4. Then decide what way to go.

?

vipentti:
I think the possible implementation will change, depending on the number of components that could actually benefit from having arbitrary event handlers.

If only a limited number of components (in addition to `Textfield`) could even use them then it might not be worth it to try and create a generic solution. Instead just do something similar to the solution in `Textfield` per component. However, if most components could benefit and work with arbitrary event handlers then a generic solution might be preferable. 

I do think in the generic case *all* components will require internal changes to accomodate the event listeners because their internal DOM varies so much there is not a singular place to put the event handlers. 

Additionally, the placement of the user provided event handlers can vary depending on the implementation and particular events. Some events `bubble` properly meaning the internal handlers could be either on the outer containers or the inner elements.

One option would be to add *all* event handlers on the 'expected' element and then use event dispatching for the events that are possibly handled by us and users (e.g. `blur` and `focus`); such as in `Menu` or in the PR for the Firefox fix #173

debois:
[...]

PR #179, opened 2016-08-15
Initial implementation of Custom Event support

vipentti:
This should resolve #173 and enable support for #164.

I have some performance concerns regarding the usage of `Dict` for keeping track of the listeners and then calling `Dict.merge` to merge internal and user provided event listeners. 

Using `Dict` was the simplest way I could come up with to allow this. However, this does limit the maximum number of Events to dispatch to 2 currently but this could be changed.

The `Dispatch` module could be separated into a separate package. 

Looking for feedback and suggestions.


OvermindDL1:
As for dict efficiency, if it matters (I am not convinced it does without benchmarks) but you could always just use a list of tuples.  The only access on them would be scanning, which is plenty fast, merging is trivial, and most people are not going to have hundreds of events, and at less than a count of 10 linked lists often outperform dicts in lookup speed as well in other languages I've used.  I do like the style from a cursory look though, need to try it out to see how it actually works.

vipentti:
Changed from Dict to List 

This change allows us to have more than 2 dispatched events
per HTML event. This is useful in cases like `Button` where
you might have a ripple and then in addition you might want to have
a tooltip

debois:
I have a thing for benchmarking things, so I went to town on the `group` implementation. This is the current implementation: 

> Benchmarking original implementation
> empty x 4,320,238 ops/sec ±1.38% (60 runs sampled)
> 1 x 918,530 ops/sec ±1.61% (60 runs sampled)
> 2 x 650,735 ops/sec ±0.98% (59 runs sampled)
> 2-overlap x 517,829 ops/sec ±1.02% (61 runs sampled)
> 5 x 102,966 ops/sec ±1.24% (60 runs sampled)
> 5-overlap-1 x 92,734 ops/sec ±0.93% (62 runs sampled)
> 5-overlap-2 x 81,728 ops/sec ±1.07% (61 runs sampled)
> 7 x 59,567 ops/sec ±1.63% (59 runs sampled)
> 10 x 48,067 ops/sec ±1.84% (56 runs sampled)
> 15 x 49,495 ops/sec ±1.85% (53 runs sampled)

I then fiddled (quite) a bit before arriving at the implementation below, which has a respectable 1-2 orders of magnitude improvement: 

> Benchmarking specialised partition / unroll-2
> empty x 51,357,133 ops/sec ±1.96% (57 runs sampled)
> 1 x 18,261,686 ops/sec ±1.71% (59 runs sampled)
> 2 x 9,330,503 ops/sec ±1.69% (59 runs sampled)
> 2-overlap x 10,529,456 ops/sec ±0.84% (63 runs sampled)
> 5 x 1,740,235 ops/sec ±1.10% (60 runs sampled)
> 5-overlap-1 x 1,626,696 ops/sec ±0.62% (62 runs sampled)
> 5-overlap-2 x 1,417,282 ops/sec ±1.55% (60 runs sampled)
> 7 x 963,474 ops/sec ±0.94% (62 runs sampled)
> 10 x 503,858 ops/sec ±0.86% (61 runs sampled)
> 15 x 503,694 ops/sec ±0.74% (63 runs sampled)

@OvermindDL1 might be right that this is totally irrelevant. I think its important nonetheless: This (and Options.collect) is the "inner loop" of elm-mdl; if people should say "but its too slow", I want to be able to say both "yeah, but virtual-dom is using 80% of the time" _and_ "we shaved off every bit of overhead we could find".

```elm


split
    : a
    -> List b
    -> List ( a, b )
    -> List ( a, b )
    -> ( List b, List ( a, b ) )
split k0 same differ xs = 
  case xs of 
    [] -> 
      (same, differ)

    ((k, v) as x) :: xs -> 
      if k == k0 then 
        split k0 (v :: same) differ xs
      else
        split k0 same (x :: differ) xs


group' : List ( a, List b ) -> List ( a, b ) -> List ( a, List b )
group' acc items = 
  case items of 
    [] -> 
      acc 

    [(k,v)] -> 
      (k, [v]) :: acc

    [(k1,v1), (k2,v2)] -> 
      if k1 == k2 then 
        (k1, [v2, v1]) :: acc
      else
        (k2, [v2]) :: (k1, [v1]) :: acc

    (k,v) :: xs -> 
      let (same, different) = 
          split k [v] [] xs
      in 
        group3' ((k, same) :: acc) different


group : List ( a, b ) -> List ( a, List b )
group = group' [] 
```

vipentti:
Awesome! I'll update the PR

OvermindDL1:
> @OvermindDL1 might be right that this is totally irrelevant.

Optimizing is fun though!  Especially if benchmarked.  ^.^

src/Material/Options/Internal.elm
```diff
@@ -12,6 +13,7 @@ type Property c m
   | Attribute (Html.Attribute m)
   | Many (List (Property c m))
   | Set (c -> c)
+  | Listener String (Json.Decode.Decoder m)
```
debois: See comments below.

src/Material/Textfield.elm
```diff
@@ -185,11 +186,8 @@ disabled =
 {-| Add custom event handlers
  -}
 on : String -> (Decoder.Decoder m) -> Property m
-on event decoder =
-    Options.set
-      (\config ->
-         { config |
-             listeners = config.listeners ++ [(Html.Events.on event decoder)]})
+on =
```
debois: Would it make sense to have also `onWithOptions`?

src/Material/Textfield.elm
```diff
@@ -282,25 +280,29 @@ defaultModel =
...
 -}
-update : Msg -> Model -> Model
+update : Msg msg -> Model -> (Model, Cmd msg)
 update action model =
   case action of
-    Input str -> 
-      { model | value = str }
-      
+    Multi msg ->
+      (model, Dispatch.forward msg)
```
debois: This should be 

    (Nothing, Dispatch.forward msg)

Unfortunately, Parts doesn't support that yet. I've opened an issue and will fix it shortly; see debois/elm-parts#10.

src/Material/Textfield.elm
```diff
@@ -349,8 +351,19 @@ view lift model options =
...
 
-    listeners =
-      config.listeners
+    -- Internal events that require special handling because user provided events
+    -- might overlap with these
+    internal =
+      [[ ("focus", Decoder.succeed (lift Focus))
+       , ("blur", Decoder.succeed (lift Blur))
+       ]
+      , (case config.value of
```
debois: I know this was my invention originally, but I get confused every time I see it. Could we leave a comment saying why there's an `input` handler iff we're given a value?

src/Material/Textfield.elm
```diff
@@ -282,25 +280,29 @@ defaultModel =
 
 {-| Component actions. `Input` carries the new value of the field.
 -}
-type Msg
+type Msg m
   = Blur
   | Focus
   | Input String
+  | Multi (Dispatch.Msg m)
```
debois: Maybe just `Dispatch` or `Forward` instead of `Multi`?

src/Material/Textfield.elm
```diff
@@ -349,8 +351,19 @@ view lift model options =
         Nothing -> []
 
 
-    listeners =
-      config.listeners
+    -- Internal events that require special handling because user provided events
+    -- might overlap with these
+    internal =
+      [[ ("focus", Decoder.succeed (lift Focus))
```
debois: Maybe add a variant of `on` to Dispatch for the very common case of not wanting to decode.  Then you'd be able to write:

    [[ ("focus", Dispatch.on1 (lift Focus))

(Not sure about the name `on1`.)

src/Material/Textfield.elm
vipentti: This is also something that could possible be changed. Do we need to be specific to have `Dispatch.Msg m`  or could we just have `Dispatch` work on any message of the type `Dispatch (List Msg)`

debois:
I made some line comments above. The big questions now are: (A) Do we apply this to all components? (B) How would we do that?

Before looking at these, a note on how to proceed: I think this will be a multi-pr project, so I propose we open a new branch `dispatch`, and merge this PR there?

I think it'd be very nice for users if there were a general mechanism for attaching event-handlers through options, so I'm in favour of applying this to all components, so I'm in favour of (A), but I'd be interested in opinions to the contrary. 

For (B), there are currently two kinds of components in elm-mdl:

1. `Html m` constructors (e.g., List, Footers, Cards, ...)
2. TEA components (e.g., Button)

For (2), we can just add the Dispatch trick to whatever messages are already there. For (1), I'm not sure what to do. In plain TEA, I can see two options: 

1. Require the user to have a Dispatch message and handle that in his own `update`.
2. Prohibit multiple subscriptions for the same event. (This requires messing with the Dispatch API; I'm not sure if it can be done transparently.)

For parts, I can only think of this:

1. Wrap the global MDL message type changing [this](https://github.com/debois/elm-mdl/blob/v7/src/Material.elm#L212-L213) to 

```elm
type Msg obs 
  = Actual (Parts.Msg Model Obs)
  | Dispatch (Parts.Msg Model Obs) 
```

and adapt `update` accordingly.

debois:
This just in: 

https://groups.google.com/forum/#!topic/elm-dev/7iUYV0KMvJc

So, we might get support for duplicate event handlers directly from elm. I propose we close this for now, then resurrect it if upstream elm doesn't come through?

vipentti:
Closing this until the fate of duplicate event handlers has been decided officially 

https://groups.google.com/forum/#!topic/elm-dev/7iUYV0KMvJc

vipentti:
Reopening this for #208 as well as due to https://groups.google.com/forum/#!topic/elm-dev/7iUYV0KMvJc not reaching any final conclusions regarding multi event handling

debois:
Agree. Could you say something about my comments above?

vipentti:
I pushed some changes to experiment with the API. If you look at 

https://github.com/vipentti/elm-mdl/blob/62f3e8f0c9ea83a2855b1639c05c0e5f111ccdd8/demo/Demo/Textfields.elm#L277-L280

and 

https://github.com/vipentti/elm-mdl/blob/62f3e8f0c9ea83a2855b1639c05c0e5f111ccdd8/src/Material/Textfield.elm#L409-L417

I think this is how I would like to have the API look. Now I had to add some extra functions to options to enable this with. I'm not sure if there is a way to avoid this. 

```elm
Options.apply' 
Options.styled'''
```

vipentti:
About how to handle these events on components, especially those without update or render, such as Card or Chip, I am not sure. The way we are doing it in `Textfield` with `Dispatch` is okay for TEA components such as `Textfield` or `Button` is alright for now but I feel like it requires quite a bit of boilerplate to remember to add it to *all* components.

debois:
I like the API suggestion. I think it'd be improved by having 
- a `on'` or `on1` for events that don't want to decode
- an `onWithOptions`, although we'd have to consider carefully whether that's safe. I think it would be once we migrate everything to use Dispatch, but we'd have to check

vipentti:
`onWithOptions` needs some extra thought on how to handle it with multi events and if it is possible even to do it

I think `Options.on'` that just succeeds is a good idea

vipentti:
Added initial support for `onWithOptions`. If users specify a listener with `onWithOptions` those set of options are applied to all the decoders for the particular event. If multiple `onWithOptions` are specified the first one is used (first one based on how listeners are grouped etc.) so its undefined behavior to have multiple `onWithOptions` for a single event

vipentti:
Added experimental support for enabling dispatch on basic view components such as `Chip`. 

Basic usage is 

```elm 

-- Step 1. Add Dispatch message 
type Msg
  = ...
  | Dispatch (Dispatch.Msg Msg)
  ...


-- Step 2. Update update accordingly 
update : Msg -> Model -> (Model, Cmd Msg)
update action model =
  case action of
    ...

    Dispatch action' ->
      (model, Dispatch.forward action' )
      
    ...
    
    
    
-- Step 3. Add `Options.dispatch Dispatch` to any component 
view : Model -> Html Msg
view model  =
    ...
    Chip.button
      [ ...
      -- Options.dispatch is REQUIRED 
      , Options.dispatch Dispatch
      
      -- If Options.dispatch is present, then multi event dispatching should just work
      , Options.on1 "click" (ChipClick index)
      , Options.on1 "click" (NoOp (toString (ChipClick index)))
      
      ...
      ]
      [ Chip.content []
          [ text value ]
      ]
```

vipentti:
Did a fairly large refactor based on @debois's suggestions of the Material.Msg and components that used to use `Parts.Msg m`. 

There is now a separate module `Material.Msg` (not sure if its needed but had some circular import issues so separated it for now). This module contains the new definition of `Material.Msg` which is 

```elm
type Msg model obs
  = Internal (Parts.Msg model obs)
  | Dispatch (Dispatch.Msg obs)
```

This simplifies the system to have messages automatically dispatched without requiring adding special `Dispatch` message per component. Instead the top level `Mdl (Material.Msg Msg)` manages everything.

Updated basic usage is 

```elm 

-- Step 1. Normally have Mdl message
type Msg
  = ...
  | Mdl (Material.Msg Msg)
  ...


-- Step 2. Have perfectly normal Mdl update
update : Msg -> Model -> (Model, Cmd Msg)
update action model =
  case action of
    ...
    Mdl action' ->
      Material.update action' model
    ...
    
    
    
-- Step 3a. Add `Options.dispatch Mdl` to any component 
view : Model -> Html Msg
view model  =
    ...
    Chip.button
      [ ...
      -- Options.dispatch is REQUIRED 
      , Options.dispatch Mdl
      
      -- If Options.dispatch is present, then multi event dispatching should just work
      , Options.on1 "click" (ChipClick index)
      , Options.on1 "click" (NoOp (toString (ChipClick index)))
      
      ...
      ]
      [ Chip.content []
          [ text value ]
      ]
      
      
-- Step 3b. Renderable components do not require adding 
-- of `Options.dispatch Mdl` as they receive it by default

      Button.render Mdl [0] model.mdl
        [ Button.colored
        , Button.ripple
        , Button.raised
        --, Button.onClick (AddChip "Amazing Chip")
        -- No need for Options.dispatch Mdl here
        -- However, you can add it and it does basically nothing
        , Options.on1 "click" (AddChip "Amazing Chip")
        , Options.on1 "click" (NoOp (toString "Amazing Chip"))
        ]
        [ text "Add chip" ]
```

OvermindDL1:
I like it, though is there no way to get rid of the `Options.dispatch Mdl`, especially since it is required, like rolling it into the components that require it, so for the Chip example above turn it into `Chip.button Mdl [..blah..]`?  Anything hard required should be in the parent call I would say.  Even if it is only required for multi-dispatch I'd still say put it on the components themselves just to make sure of no surprises by accidentally forgetting it somewhere.

vipentti:
I think adding constructors that take the `Mdl` as parameter is probably reasonable. I did not put it there initially because I'm still hoping for Native multi event support and if that ever arrives you could just make `Options.dispatch` into a `None` and possibly avoid MAJOR version bump

OvermindDL1:
> I think adding constructors that take the Mdl as parameter is probably reasonable. I did not put it there initially because I'm still hoping for Native multi event support and if that ever arrives you could just make Options.dispatch into a None and possibly avoid MAJOR version bump

Ah, not a bad idea with that yeah.  Still means it is easy to forget however?

vipentti:
It's possible to forget, currently we support singular events if no lift has been provided. But if multiple events have been provided using `Options.on` variants with no `Options.dispatch` we just call `Debug.crash` and explode for now

vipentti:
`Material.update` now takes the outer update function as the
first argument. This allows us to call `Dispatch.update`
with the outer update when dispatching multiple events.

The reason for `Dispatch.update` is performance. `Dispatch.forward`
would cause several re-renders when dispatching commands where as
`Dispatch.update` will run through all the messages and only after
its done, we return the updated model. In addition `Dispatch.update`
should properly dispatch any commands from the dispatched messages.

demo/Demo/Textfields.elm
```diff
@@ -81,41 +82,59 @@ pure =
...
     Slider value ->
-      { model | length = value } |> pure
+      { model | length = value } ! []
 
     SetFocus5 x ->
-      { model | focus5 = x } |> pure
+      { model | focus5 = x } ! []
 
     SelectionChanged selection ->
+      -- High-frequency event; return referentially equal model on NOP.
```
debois: This is dead code, yes? Maybe arrange `Msg` so that you don't have that. (Maybe an outer message featuring HF and Regular events, `update'` dispatches only Regular?)

src/Dispatch.elm
```diff
@@ -0,0 +1,258 @@
...
+map2nd : (b -> c) -> (a,b) -> (a,c)
+map2nd f (x,y) = (x, f y)
+
+
+{-| Runs batch update
+-}
+update : (a -> b -> ( b, Cmd a )) -> Msg a -> b -> (b, Cmd a)
+update update (Forward msg) model =
+  let
+    inner cmd (m, gs) =
```
debois: Slightly sleeker:
```elm
inner msg (model, gs) = 
  update msg model 
    |> map2nd (flip (::) gs)
```

src/Dispatch.elm
```diff
@@ -0,0 +1,258 @@
...
+
+
+{-| Applies given decoders to the same initial value
+   and return the applied results as a list
+-}
+applyMultipleDecoders : List (Json.Decoder m) -> Json.Decoder (List m)
+applyMultipleDecoders decoders =
+  let
+    processDecoder initial decoder =
+      case (Json.decodeValue decoder initial) of
```
debois:
```elm
processDecoder initial decoder = 
  Json.decodeValue decoder initial |> Result.toMaybe
```

src/Dispatch.elm
```diff
@@ -0,0 +1,258 @@
...
+      inner
+      (model, [])
+      msg
+    |> map2nd Cmd.batch
+
+
+{-| Applies given decoders to the same initial value
+   and return the applied results as a list
+-}
+applyMultipleDecoders : List (Json.Decoder m) -> Json.Decoder (List m)
```
debois: Neat.

src/Dispatch.elm
```diff
@@ -0,0 +1,258 @@
...
+    -- NOTE: This need to be changed, currently only for debugging
+    x :: xs ->
+      Debug.crash <| "Multiple decoders for Event '" ++ event ++  "' with no `Options.dispatch Mdl`"
+      -- Html.Events.onWithOptions event options x
+      --   |> Just
+
+
+{-| Updates value by given function if found, inserts otherwise
+-}
+upsert : comparable -> m -> (Maybe m -> Maybe m) -> Dict comparable m -> Dict comparable m
```
debois: Dead code.

src/Dispatch.elm
```diff
@@ -0,0 +1,258 @@
...
+    [] ->
+      Nothing
+
+    [ x ] ->
+      Html.Events.onWithOptions event options x
+        |> Just
+
+    -- NOTE: This need to be changed, currently only for debugging
+    x :: xs ->
+      Debug.crash <| "Multiple decoders for Event '" ++ event ++  "' with no `Options.dispatch Mdl`"
```
debois: Help me understand how we arrange that this clause is reachable iff `Options.dispatch Mdl` was not used?

src/Dispatch.elm
```diff
@@ -0,0 +1,258 @@
...
+    |> List.filterMap identity
+    |> List.head
+    |> Maybe.withDefault Html.Events.defaultOptions
+
+
+{-| Combines decoders for events and returns event listeners
+-}
+listeners :
+  (Msg a -> a)
+  -> List ( String, List (Pair a) )
```
debois: What's `Pair a`?

src/Dispatch.elm
```diff
@@ -0,0 +1,258 @@
...
+
+
+
+{-| Message type
+-}
+type Msg m
+  = Forward (List m)
+
+
+type alias Pair m =
```
debois: The name `Pair` is not so helpful. Maybe just `Decoder m`?

src/Material/Button.elm
```diff
@@ -228,26 +225,29 @@ view lift model config html =
...
     summary = Options.collect defaultConfig config
 
-    startListeners =
-      if summary.config.ripple then
-        [ Ripple.downOn' lift "mousedown" |> Just
-        , Ripple.downOn' lift "touchstart" |> Just
+
+    listeners =
+      Options.many
+        [ Ripple.downOn_ lift "mousedown"
```
debois: What's with the "..._"?

src/Material/Button.elm
```diff
@@ -228,26 +225,29 @@ view lift model config html =
...
+        , Ripple.upOn_ lift "mouseleave"
         ]
-      else
-        []
+
 
     stopListeners =
       let handle =
-        Just << if summary.config.ripple then blurAndForward else Helpers.blurOn
+            Just << Helpers.blurOn
```
debois: You got rid of the blurAndForward terribleness!

src/Material/Button.elm
```diff
@@ -263,18 +263,18 @@ view lift model config html =
...
       (if summary.config.ripple then
           List.concat
             [ html
             -- Ripple element must be last or blurAndForward hack fails.
             , [ Html.App.map lift <| Ripple.view'
                   [ class "mdl-button__ripple-container"
-                  --, Helpers.blurOn "mouseup"
-                  , Ripple.upOn "blur"
-                  , Ripple.upOn "touchcancel"
+                  -- , Ripple.upOn "blur"
```
debois: Could you help me understand how Ripple changed?

src/Material/Button.elm
```diff
@@ -396,6 +396,10 @@ type alias Container c =
   { c | button : Indexed Model }
 
 
+view' : (Msg.Msg (Container c) m -> m) -> (Msg -> m) -> Model -> List (Property m) -> List (Html m) -> Html m
```
debois: Help me understand this?

src/Material/Icon.elm
```diff
@@ -51,8 +50,7 @@ with `Button.icon` instead.)
 -}
 onClick : m -> Property m
 onClick x =
-  Options.set
-    (\config -> { config | onClick = Just (Html.Events.onClick x) })
+  Options.on1 "click" x
```
debois: Maybe just give up and implement a general `onClick` in `Options`.

src/Material/Options.elm
```diff
@@ -85,6 +99,8 @@ type alias Summary c m =
   , css : List (String, String)  
   , attrs : List (Attribute m)
   , config : c
+  , listeners : List (String, (Decoder.Decoder m, Maybe Html.Events.Options))
```
debois: Cleaner:
```elm
, listeners : List (String, Dispatch.Pair m)
```
Even cleaner if `Pair m` -> `Decoder m`.

src/Material/Options.elm
```diff
@@ -85,6 +99,8 @@ type alias Summary c m =
...
+  , dispatch : Maybe (Dispatch.Msg m -> m)
```
debois: Clever.

src/Material/Button.elm
vipentti: It did not but I am unable to have it break with just having the button events trigger `Ripple.upOn`

src/Material/Button.elm
vipentti: Just a temporary name

src/Material/Button.elm
vipentti: That is dead code, it was meant to just take care of injecting `Options.dispatch` but there is a generic version of it currently residing in `Options.inject`

src/Dispatch.elm
vipentti: Basically `onSingle` is only called if you have not provided `Options.dispatch` to your elements while using `Options.on`. Meaning either you forgot it hence `Debug.crash` for now in the last clause, or its fine with a single event

src/Material/Options.elm
```diff
@@ -347,3 +375,72 @@ For example `Textfield`:
...
+{-| Add custom event handlers with options
+ -}
+onWithOptions : String -> Html.Events.Options -> (Decoder.Decoder m) -> Property c m
+onWithOptions evt options =
+  Listener evt (Just options)
+
+
+{-| Add a lifting function that is **required** for multi event dispatch.
+
+**NOTE** If this is missing no events are dispatched with `Options.on`
```
debois: Is this true? I'd have thought you just don't get dispatch of multiple handlers for the same event.

src/Material/Options.elm
```diff
@@ -347,3 +375,72 @@ For example `Textfield`:
...
+ -}
+dispatch : (Msg.Msg a m -> m) -> Property c m
+dispatch lift =
+  Lift (Msg.Dispatch >> lift)
+
+
+{-| Add a lifting function that is **required** for multi event dispatch.
+
+**NOTE** If this is missing no events are dispatched with `Options.on`
```
debois: Again.

src/Material/Options.elm
```diff
@@ -347,3 +375,72 @@ For example `Textfield`:
...
+ -}
+dispatch' : (Dispatch.Msg b -> b) -> Property c b
+dispatch' =
+  Lift
+
+
+{-| Inject dispatch
```
debois: Little more help in the comment would be nice :)

src/Material/Ripple.elm
```diff
@@ -142,6 +143,16 @@ downOn' f name =
   Html.Events.on name (Json.map (Down >> f) geometryDecoder)
 
 
+downOn_ : (Msg -> m) -> String -> Options.Property c m
```
debois: Not crazy about the `*_` suffix.

src/Material/Table.elm
```diff
@@ -248,9 +246,10 @@ type Order
 
 {-| Dispatch given message when cell is clicked. 
 -}
-onClick : m -> Property { a | onClick : Maybe (Attribute m) } m
+onClick : m -> Property c m
```
debois: Again, we should consider getting rid of all these and make `Options.onClick` the canonical thing you use.

src/Material/Msg.elm
```diff
@@ -0,0 +1,8 @@
+module Material.Msg exposing (..)
+
+import Dispatch
+import Parts
+
+type Msg model obs
+  = Internal (Parts.Msg model obs)
+  | Dispatch (Dispatch.Msg obs)
```
debois: Could this be 
```elm
  | Dispatch (List obs)
```
?

src/Material/Options.elm
vipentti: This is actually false, need to update the comments accordingly

src/Material/Msg.elm
vipentti: It could be

src/Dispatch.elm
```diff
@@ -0,0 +1,262 @@
...
+map2nd f ( x, y ) =
+  ( x, f y )
+
+
+{-| Runs batch update
+-}
+update : (a -> b -> ( b, Cmd c )) -> Msg a -> b -> ( b, Cmd c )
+update update (msg) model =
+  let
+    inner cmd ( m, gs ) =
```
debois: Consider lifting out of `update`. (Maybe. I can't figure out how bad `let` are for perf.)

src/Dispatch.elm
```diff
@@ -0,0 +1,262 @@
...
+-}
+applyMultipleDecoders : List (Json.Decoder m) -> Json.Decoder (List m)
+applyMultipleDecoders decoders =
+  let
+    processDecoder initial decoder =
+      Json.decodeValue decoder initial
+        |> Result.toMaybe
+  in
+    Json.customDecoder Json.value
+      (\initial ->
```
debois: Lift the closure out of applyMultipleDecoders. (Won't affect the benchmark, which never tries multiple-event dispatching.)

src/Dispatch.elm
```diff
@@ -0,0 +1,262 @@
...
+      let
+        -- NOTE: This probably needs to be changed, currently only for debugging
+        _ = Debug.log "WARNING" ("Multiple decoders for Event '" ++ event ++ "' with no `Options.dispatch Mdl`")
+      in
+        Html.Events.onWithOptions event options x
+
+
+pickOptions : List (Decoder a) -> Html.Events.Options
+pickOptions decoders =
+  List.map snd decoders
```
debois:
```
pickOptions : List (Decoder a) -> Html.Events.Options
pickOptions =
    List.filterMap snd 
    >> List.head
    >> Maybe.withDefault Html.Events.defaultOptions
```
Elm may like the eta-expanded version better, i.e.,
```
pickOptions : List (Decoder a) -> Html.Events.Options
pickOptions decoders =
    List.filterMap snd decoders
    |> List.head
    |> Maybe.withDefault Html.Events.defaultOptions
```

debois: (I think this will likely make a difference.)

src/Dispatch.elm
```diff
@@ -0,0 +1,262 @@
...
+-}
+onSingle :
+  String
+  -> Html.Events.Options
+  -> List (Json.Decoder msg)
+  -> (Html.Attribute msg)
+onSingle event options decoders =
+  case decoders of
+    [] ->
+      noop
```
debois: Is this better than returning `Maybe (Html.Attribtue msg)` and using `filterMap` as in the benchmark you used previously? I'd think the benchmark was better. 

Watch out, though: You sometimes do
```
   List.map f xs |> List.filterMap identity
```
which processes the list twice; better to do directly
```
   List.filterMap f xs 
```
which is equivalent.

src/Dispatch.elm
```diff
@@ -0,0 +1,262 @@
...
+If a lifting argument is available use
+`Dispatch.listeners` instead.
+-}
+listeners' :
+  List ( String, List (Decoder a) )
+  -> List (Html.Attribute a)
+listeners' items =
+  items
+    |> List.map
+        (\( event, decoders ) ->
```
debois: Lose the closure—change the signature of `onSingle` to take `(event, decoders)` directly; do `pickOption` itself, and avoid the `List.map fst decoders` below by applying `fst` inside `onSingle`. 

Also consider trying specialising `onSingle` for the common case of one event-one decoder.
 

(I think this'll help.)

src/Material/Options.elm
```diff
@@ -135,24 +175,44 @@ collect1' options acc =
...
+    that way internal class and specific attributes can override those
+    provided by the user
+     -}
+    List.append
+      summary.attrs
+      (  Html.Attributes.style summary.css
+      :: Html.Attributes.class (String.join " " summary.classes)
+      :: attrs
+      )
+      |> (flip (++) summary.internal)
```
debois: The expression `A ++ B` is O(n) in the size of `A`, i.e., it traverses the `A` list, when it reaches the end, it adds `B`. The `List.append` is in principle the same as `++`, but in Elm, it turns out `(++)` is a [native implementation](https://github.com/elm-lang/core/blob/4.0.5/src/Native/Utils.js#L246-L270) and `List.append` is [implemented](https://github.com/elm-lang/core/blob/4.0.5/src/List.elm#L286-L295) in terms of `fold`.

It might make a difference if you used either `List.concat` or put `++` between each list here.

src/Material/Options.elm
```diff
@@ -135,24 +175,44 @@ collect1' options acc =
     Class x -> { acc | classes = x :: acc.classes }
     CSS x -> { acc | css = x :: acc.css }
     Attribute x -> { acc | attrs = x :: acc.attrs }
+    Internal x -> { acc | internal = x :: acc.internal }
     Many options -> List.foldl collect1' acc options
     Set _ -> acc 
+    Listener event options decoder ->
```
debois: Ordering might matter—try put higher-frequency cases first. Also, maybe tuck away Set _ and nop into a final `_ -> acc` case.

src/Material/Options.elm
```diff
@@ -135,24 +175,44 @@ collect1' options acc =
...
 addAttributes : Summary c m -> List (Attribute m) -> List (Attribute m)
-addAttributes summary attrs = 
-  List.append
-    attrs
-    (  Html.Attributes.style summary.css 
-    :: Html.Attributes.class (String.join " " summary.classes) 
-    :: summary.attrs
-    )
+addAttributes summary attrs =
+  let
```
debois: Again, maybe check what comes out in elm.js when you use let—I seem to recall elm generates a closure because it doesn't want to rely on whatever version of javascript it is that gives us a let construct.

src/Material/Options.elm
vipentti: `++` it is

FAQ.md
```diff
@@ -126,7 +126,7 @@ The 7.0.0 release changes the required boilerplate in two aspects.
           case message of
             ...
             Mdl message' ->
-              Material.update message' model
+              Material.update update message' model
```
knewter: This can't be right can it?

README.md
```diff
@@ -76,7 +76,7 @@ The 7.0.0 release changes the required boilerplate in two aspects.
           case message of
             ...
             Mdl message' ->
-              Material.update message' model
+              Material.update update message' model
```
knewter: Or this?  Looking at the `Material.update` function it seems to not be the case that it changed this way...

demo/Demo/Textfields.elm
```diff
@@ -76,45 +81,57 @@ selectionDecoder =
...
+          |> Just
+
+    Focus -> 
+      ( model 
+      , Task.perform (always Nop) (always Nop) (Dom.focus "my-textfield")
+      ) |> Just
+
     Mdl action' ->
-      Material.update action' model |> Just
+      Material.update action' model
```
knewter: `msg`

PR #180, opened 2016-08-15
Docs/lists to list

domenkozar:
Renames

```
import Material.List as Lists
```
to
```
import Material.List as List
```

Now it's consistent with the rest of the demos. Note that I haven't renamed the file itself.

cc @debois

### DIFF
--- a/demo/Demo/Lists.elm
+++ b/demo/Demo/Lists.elm
@@ -5,7 +5,7 @@ import Html exposing (..)
 import Set exposing (Set)
 import String
 
-import Material.List as Lists
+import Material.List as List
 import Material
 import Material.Options as Options exposing (when, css)
 import Material.Icon as Icon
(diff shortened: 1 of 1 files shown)


debois:
Thanks, really appreciate it!

I'm a bit concerned there'll be confusion with `List` from core, though—this is why I used `Lists` in the first place. But maybe I'm overly cautious. @aforemny, @vipentti, any thoughts?

aforemny:
Given that fact that we are importing `module Material.Lists as Lists` I think it is the right way to go.

OvermindDL1:
@aforemny Are there any name clashes though?

Remember that if you import multiple modules into the same name in Elm it all 'just works' because each name in that module is unique.  If you have multiple functions with the same name then the compiler will yell at you only if that function is used, at which point you just have to be explicit to the module it belongs to.  I don't think Material.List has or could even have any overlapping function names with the stock List module.

debois:
Since the Demo is used quite often by sometimes newbie users as guidance, conceivably someone will at some point be confused about what `List` means exactly. For that reason, I'm closing this without merging. 

@domenkozar: I'm really sorry we couldn't take this—I do hope you'll keep them coming regardless!

domenkozar:
@debois no worries. I'll leave the branch alive if this comes up again to be reconsidered. Thanks!

PR #186, opened 2016-08-16
Removed Phoenix Elm Chat based on slack discussion

hakonrossebo:
Will add this project when it is ready.


hakonrossebo:
Based on the last comment from Josh Adams, we don't need to merge this PR

debois:
Ok, closing.

Issue #165, opened 2016-08-10
Allow for something other than MDL icons for the menu button.


Lattyware:
I am currently porting a project ([massivedecks](https://github.com/Lattyware/massivedecks)) to use elm-mdl (from muicss integrated manually) - we used fontawesome for icons in the past, and the material design icon set is missing a number of icons used, so I have stuck with fontawesome and simply not included the MDL icons - that all seems fine and is working well, except the menu, which appears to only have support for MDL icons for the menu button.

It would unblock us to allow arbitrary content for the menu button as a replacement for the icon (I can also see this holding value for those that want text instead of an icon, for example). Otherwise we have to include two icon fonts, when one would do fine.

OvermindDL1:
As a work-around for now in your site CSS you can override the those to make them invisible and add an :after element that adds whatever you need for a font (via some parent div with, say, a `data-myicon="somefontawesome"` value or so, can set it with something like:
```css
some[data-myicon] > selector {
  make it invisible however you want, like width=0px should work and put the :after in the right place
}

some[data-myicon] > selector:after {
  content: attr(data-badge);
}
```
I doubt that is the right setting or if the attr works from a parent, but there is always some way to do it via css.

debois:
@Lattyware: I think this is likely something we want to do, but a little time might pass before anyone gets around to it. (We'd take a PR, of course.) 

In the short term, is @OvermindDL1's workaround an option for you?

debois:
@Lattyware: This is something we'd like to do, but a little time might pass before anyone gets around to it. (We'd take a PR, of course.) 

In the short term, is @OvermindDL1's workaround an option for you?

@aforemny: Is this something you'd be interested in looking at?

aforemny:
@debois Yes I am! :)

aforemny:
In the mean-time, [here](https://gist.github.com/aforemny/0de30b52978fbb61a079071d31b76d48) is a more elaborate example of @OvermindDL1's work-around.

aforemny:
When allowing menus to be opened by arbitrary elements, the tooltip implementation comes to mind. Menus are different from tooltips in that on opening they are decoding information from the element that opens them _as well_ as the menu itself.

[...]

PR #190, opened 2016-08-16
Initial implementation Chip component

### LINKED ISSUE #128: Implement Chip component
Apparently underway in upstream: https://github.com/google/material-design-lite/tree/mdl-1.x/src/chip.

vipentti:
Initial implementation of the `Chip` component new in version 1.2.0 to resolve #128 

Demo could use some beefing up perhaps.

### DIFF
--- a/demo/Demo/Chips.elm
+++ b/demo/Demo/Chips.elm
@@ -129,42 +129,42 @@ demoContainer (html, code) =
 
 demoChips : List (Grid.Cell m)
 demoChips =
-  [ ( Chip.chipSpan []
+  [ ( Chip.span []
         [ Chip.content []
             [ text "Basic Chip" ]
         ]
     , """
-      Chip.chipSpan []
+      Chip.span []
         [ Chip.content []
             [ text "Basic Chip" ]
         ]
        """
     )
-  , ( Chip.chipSpan
+  , ( Chip.span
         [ Chip.deleteIcon "cancel" ]
         [ Chip.content []
             [ text "Deletable Chip" ]
         ]
     , """
-      Chip.chipSpan
+      Chip.span
         [ Chip.deleteIcon "cancel" ]
         [ Chip.content []
             [ text "Deletable Chip" ]
         ]
        """
     )
-  , ( Chip.chipButton []
+  , ( Chip.button []
         [ Chip.content []
             [ text "Button Chip" ]
         ]
     , """
-      Chip.chipButton []
+      Chip.button []
         [ Chip.content []
             [ text "Button Chip" ]
         ]
        """
     )
-  ,  ( Chip.chipSpan []
+  ,  ( Chip.span []
           [ Chip.contact Html.span
               [ Color.background Color.primary
               , Color.text Color.white
@@ -174,7 +174,7 @@ demoChips =
               [ text "Contact Chip" ]
           ]
      , """
-        Chip.chipSpan []
+        Chip.span []
           [ Chip.contact Html.span
               [ Color.background Color.primary
               , Color.text Color.white
@@ -185,7 +185,7 @@ demoChips =
           ]
         """
      )
-  , ( Chip.chipSpan
+  , ( Chip.span
         [ Chip.deleteLink "#chips" ]
         [ Chip.contact Html.span
             [ Color.background Color.primary
@@ -196,7 +196,7 @@ demoChips =
             "Deletable Contact Chip"
         ]
     , """
-      Chip.chipSpan
+      Chip.span
         [ Chip.deleteLink "#chips"
         ]
         [ Chip.contact Html.span
@@ -209,7 +209,7 @@ demoChips =
         ]
        """
     )
-  , ( Chip.chipSpan []
+  , ( Chip.span []
         [ Chip.contact Html.img
             [ Options.css "background-image" "url('assets/images/elm.png')"
             , Options.css "background-size" "cover"
@@ -219,7 +219,7 @@ demoChips =
             [ text "Chip with image" ]
         ]
     , """
-      Chip.chipSpan []
+      Chip.span []
         [ Chip.contact Html.img
             [ Options.css "background-image"
                 "url('assets/images/elm.png')"
@@ -298,7 +298,7 @@ view model  =
                   ]
                   ((Dict.toList model.chips
                    |> List.map (\ (index, value) ->
-                                  Chip.chipButton
+                                  Chip.button
                                     [ Options.css "margin" "5px 5px"
                                     , Chip.onClick (ChipClick index)
                                     , Chip.deleteClick (RemoveChip index)
@@ -322,7 +322,7 @@ view model  =
       , Code.code
           [ Options.css "margin-top" "20px"]
           """
-            Chip.chipButton
+            Chip.button
               [ Options.css "margin" "5px 5px"
               , Chip.onClick (ChipClick index)
               , Chip.deleteClick (RemoveChip index)
--- a/demo/Demo/Chips.elm
+++ b/demo/Demo/Chips.elm
@@ -47,7 +47,7 @@ update action model =
 demoContainer : (Html m, String) -> (Grid.Cell m)
 demoContainer (html, code) =
   Grid.cell
-  [ Grid.size Grid.All 4 ]
+  [ Grid.size Grid.All 6 ]
   [ Options.div
       [ css "text-align" "center" ]
       [ html ]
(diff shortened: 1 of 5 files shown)


demo/Demo/Chips.elm
```diff
@@ -0,0 +1,205 @@
...
+
+        , Chip.button []
+            [ Icon.view "cancel" [] ]
+        ]
+    , """
+      Chip.chip
+        [ Chip.deletable ]
+        [ Chip.text []
+            [ text "Deletable Chip" ]
+
```
debois: Unnecessary linebreak? (Also above.)

demo/Demo/Chips.elm
```diff
@@ -0,0 +1,205 @@
...
+        , Chip.button []
+            [ Icon.view "cancel" [] ]
+        ]
+       """
+    )
+  , ( Chip.chip' Html.button
```
debois: Neat.

demo/Demo/Chips.elm
```diff
@@ -0,0 +1,205 @@
...
+        [ Options.type' "button" ]
+        [ Chip.text []
+            [ text "Button Chip" ]
+        ]
+       """
+    )
+  ,  ( Chip.chip
+          [ Chip.contact ]
+          [ Chip.contactItem
+              [ Options.cs "mdl-color--teal"
```
debois: Would this work?

```elm
[ Color.background Color.primary
, Color.text Color.white 
] 
```

demo/Demo/Chips.elm
```diff
@@ -0,0 +1,205 @@
...
+              ]
+              [ text "A" ]
+          , Chip.text []
+              [ text "Contact Chip" ]
+          ]
+     , """
+        Chip.chip
+          [ Chip.contact ]
+          [ Chip.contactItem
+              [ Options.cs "mdl-color--teal"
```
debois: See above.

demo/Demo/Chips.elm
```diff
@@ -0,0 +1,205 @@
...
+              [ text "Contact Chip" ]
+          ]
+        """
+     )
+  , ( Chip.chip
+        [ Chip.contact
+        , Chip.deletable
+        ]
+        [ Chip.contactItem
+            [ Options.cs "mdl-color--teal"
```
debois: See above.

demo/Demo/Chips.elm
```diff
@@ -0,0 +1,205 @@
...
+            [ Options.attribute <| Html.href "#chips" ]
+            [ Icon.view "cancel" [] ]
+        ]
+    , """
+      Chip.chip
+        [ Chip.contact
+        , Chip.deletable
+        ]
+        [ Chip.contactItem
+            [ Options.cs "mdl-color--teal"
```
debois: See above.

src/Material/Chip.elm
```diff
@@ -0,0 +1,123 @@
...
+{-| Create a chip contained in the given element
+-}
+chip' : HtmlElement msg -> List (Options.Property c msg) -> List (Html msg) -> Html msg
+chip' element props content =
+  Options.styled element
+    (cs "mdl-chip" :: props)
+    content
+
+
+{-| Creates a chip using `Html.span`
```
debois: Neat.

debois:
I like this. The API looks very nice when used in the demo.

Could you say a little bit about why Chips are implemented as element transformers and why we need to expose `chips'`?

It seems Chips could have been implemented as just a bunch of `Property c m`, which users would then have to apply correctly. (Ok, maybe that's why?)

vipentti:
@debois I initially had them as properties but I felt the boilerplate to create a chip was too much, especially because applying them correctly would require knowledge about the structure. 

e.g. 
```elm
Options.styled Html.span
  [ Chip.chip
  , Chip.deletable
  ]
  [ Options.styled Html.span
      [ Chip.text ]
      [ text "Deletable Chip" ]
  , Options.styled' Html.button
    [ Chip.action ]
    [ Html.type' "button" ]
    [ Icon.view "cancel"
        []
    ]
  ]
```

vs 

```elm
Chip.chip
  [ Chip.deletable ]
  [ Chip.text []
      [ text "Deletable Chip" ]
  
  , Chip.button []
      [ Icon.view "cancel" [] ]
  ]
```

Also the reason I expose `chip'` and `text'` and other variants is because `Chip` is typically a `span` or `div` but it could also be a `button` and such.

OvermindDL1:
Why not have Chip.chip take a node as the first element?  That way we could do:
```elm
Chip.chip Html.div
  [ Chip.deletable ]
  [ Chip.text [] {- I'd keep Chip.text for ease of repeatability -}
      [ text "Deletable Chip" ]
  , Chip.action Html.Button [] {- Or maybe Material.Button, or Html.href or whatever? -}
      [ Icon.view "cancel" [] ]
  ]
```

vipentti:
@OvermindDL1 That's what `Chip.chip'` allows, we could always swap those around if needed

OvermindDL1:
Ah, I missed that (was concentrating on other code >.>).  Looks good in that case although personally I'd rename `chip'` to `chip` and rename `chip` to `chip` to `chipSpan` or so.  I always prefer explicit over implicit in naming.  :-)

debois:
Couple of questions: 

1. Can we get rid of `chip.deletable`? From the [upstream implementation](https://github.com/google/material-design-lite/blob/mdl-1.x/src/chip/_chip.scss#L81-L83) what it does is add padding to accommodate an icon at the right end of the chip; can we somehow roll that into the `...action` bits?

1a. Otherwise, we should probably give it a different name, since it's necessary whenever your chip has an icon, yes?

2. According to [the spec](https://material.google.com/components/chips.html#chips-usage), clicking a chip should open "a full detailed view (either in a card or full screen) or a menu of options". We can talk about `menu` when @aforemny finalises #165, but for now, is there a way for me to add an action to clicking the entire chip? If there is, can we show that in the demo?

3. I'm thinking it maybe isn't necessary to let user's customise elements used in actions (typically delete). We just need to make sure he has the option of either (a) dispatching a message, or (b) going to some URL. Is this possible?

4. I'm unsure why it's necessary to let user's decide what the top-level element of a chip should be. I can see `button` and `span` renders differently, but beyond those two options, are there other things he may want?

I was thinking about releasing a patch release shortly, but hold off Chips until we get the API _just_ right?

vipentti:
I think we can hold off `Chips` until we finalize the API

vipentti:
Regarding 3. I think this is where we can decide how much we want to simply follow `MDL` and how much we want to be an UI implementation that uses MDL.  

MDL says: 
> Create a container element for the chip; typically `<span>` and `<div>`
> are used, but any container element should work equally well. If you need
> interactivity, use a `<button>`, or add the tabindex attribute to your container.

I will work on adding an interactive example(s) with deletable actions and such.

vipentti:
Did a refactor of the Chip API. It does still look somewhat the same to the user but internally there are more changes.

1. Added opaque `Content` type to `Chip` to only limit certain types of content that can be added (more strict than `MDL` but closer to the spec).

3. `deletable` is no longer a property or an action that can be provided by the user.
To make a chip "deletable" simply add one of the following to the Chip: `Chip.deleteIcon`, `Chip.deleteClick`, `Chip.deleteAction`. Adding one of these will automatically add the `mdl-chip--deletable` class as well as the icon. Icon can be changed by `Chip.deleteIcon`. By default `Icon.view "cancel" []` is used.

4. `contact` is no longer a property. It now is a function that creates a contact-element, and if such an element is present then the `mdl-chip--contact` class is added to the chip automatically. 

5. The naming of the Chip functions are still in the open. I'm not sure if the `chip` prefix is necessary for the `Chip.chipSpan` and `Chip.chipButton` types. The could probably just be named `Chip.span` and `Chip.button` (also to discourage someone just exposing everything).

6. Chips also support custom event handlers with `Chip.on`. There is really no particular reason for it.

Some examples: 
```elm
Chip.chipSpan
  [ Chip.deleteIcon "cancel" ]
  [ Chip.content []
      [ text "Deletable Chip" ]
  ]

Chip.chipSpan
  [ Chip.deleteLink "#chips"
  ]
  [ Chip.contact Html.span
      [ Color.background Color.primary
      , Color.text Color.white
      ]
      [ text "A" ]
  , Chip.content []
      [ text "Deletable Contact Chip" ]
  ]

Chip.chipButton
  [ Chip.onClick (ChipClick index)
  , Chip.deleteClick (RemoveChip index)
  ]
  [ Chip.content []
      [ text ("Amazing Chip " ++ toString index ) ]
  ]
```


As always looking for feedback and suggestions.

OvermindDL1:
+1 vote for removing the `chip` prefix to become `Chip.span` and `Chip.button` and such as well.


Is there any chance for this Chip container thing that is listed in the animation at https://material.google.com/components/chips.html#chips-behavior so we could, for example, make adding users, tags, etc... like how it is in GMail, type a name in a textbox, auto-completes out and it resolves into a Chip?


As per https://material.google.com/components/chips.html#chips-contact-chips what is the way to pop-up the extra details box for a user (such as in the linked example of being able to choose from a few different email addresses via the Material defined Normal, Focused, Pressed, and Activated states (are able to active multiple in some implementations I've seen as well).

vipentti:
The current upstream MDL does not implement the full Contact chip specification; missing Open contact chip. It might be possible to implement it with Lists  and some absolutely positioned container. Something like how tooltip works. 

As for the resolving to Chips, it should be doable just having a container with the chips and a textfield.

OvermindDL1:
> The current upstream MDL does not implement the full Contact chip specification; missing Open contact chip. It might be possible to implement it with Lists and some absolutely positioned container. Something like how tooltip works.

I keep forgetting about 'mdl', I defer to the spec itself as mdl is not always compliant, and I prefer compliance...  >.>

> As for the resolving to Chips, it should be doable just having a container with the chips and a textfield.

Yeah just an easy form that is to be repeated often, would be useful to have that as another new component (CardContainer?) perhaps someday.

PR #199, opened 2016-08-20
Add support for expandable textfields

### LINKED ISSUE #198: Textfield: Implement Expandable Textfield
https://github.com/google/material-design-lite/tree/mdl-1.x/src/textfield#to-include-an-expandable-mdl-text-field-component

and 

https://getmdl.io/components/index.html#textfields-section

vipentti:
Initial support to resolve #198.  

Currently requires user to provide ID for the element as a parameter to `Textfield.expandable` for example `Textfield.expandable "expandable-textfield-id"`. This allows the implementation to stay extremely simple and not require any sort of JavaScript or other hacks to achieve the necessary functionality.

### DIFF
--- a/demo/Demo/Textfields.elm
+++ b/demo/Demo/Textfields.elm
@@ -224,6 +224,22 @@ textfields model =
         ]
        """
     )
+  , ( "Expandable textfield"
+    , Textfield.render Mdl [11] model.mdl
+        [ Textfield.label "Expandable"
+        , Textfield.floatingLabel
+        , Textfield.expandable "expandable-1"
+        , Textfield.expandableIcon "search"
+        ]
+    , """
+      Textfield.render Mdl [11] model.mdl
+        [ Textfield.label "Expandable"
+        , Textfield.floatingLabel
+        , Textfield.expandable "expandable-1"
+        , Textfield.expandableIcon "search"
+        ]
+       """
+    )
   , ( "Multi-line textfield"
     , Textfield.render Mdl [6] model.mdl
         [ Textfield.label "Default multiline textfield"
--- a/demo/Demo/Textfields.elm
+++ b/demo/Demo/Textfields.elm
@@ -229,24 +229,14 @@ textfields model =
     , Textfield.render Mdl [11] model.mdl
         [ Textfield.label "Expandable"
         , Textfield.floatingLabel
-        -- NOTE: The id's must match
-        -- between Html.Attributes.id and Textfield.expandable
-        , Options.inner
-            [ Options.attribute <| Html.Attributes.id "expandable-1"
-            ]
-        , Textfield.expandable "expandable-1"
+        , Textfield.expandable "id-of-expandable-1"
         , Textfield.expandableIcon "search"
         ]
     , """
       Textfield.render Mdl [11] model.mdl
         [ Textfield.label "Expandable"
         , Textfield.floatingLabel
-        -- NOTE: The id's must match
-        -- between Html.Attributes.id and Textfield.expandable
-        , Options.inner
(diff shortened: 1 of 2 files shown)


vipentti:
This PR might need some minor changes based on #206. Or at least a documentation update to properly document that if users provide an `Options.inner [ Options.id "id-match" ]` then that id has to match with `Textfield.expandable "id-match"` otherwise the results might be unexpected.

debois:
In `Toggles`, I avoided the need for ids by simply not adding the `for` attributes to labels—it seems modern browsers associate labels correctly by proximity. Is that an option here?

debois:
Also, could we note someplace (maybe in a comment in the code, maybe in an issue here) that we have to go back and fix this once we get proper generation of Ids for use with the DOM API?

vipentti:
Unfortunately due to the structure of the expandable textfield, leaving `for` attribute seems to stop it from working

demo/Demo/Textfields.elm
```diff
@@ -224,6 +225,32 @@ textfields model =
         ]
        """
     )
+  , ( "Expandable textfield"
+    , Textfield.render Mdl [11] model.mdl
+        [ Textfield.label "Expandable"
+        , Textfield.floatingLabel
+        -- NOTE: The id's must match
```
debois: Can't we make a single property that sets both? So that you can't accidentally forget one or mismatch them?

E.g., `expandable` overrides the id set in inner?

vipentti: Expandable actually sets the id but if users for some reason provide `Options.inner` with ID then those ids must match

vipentti: I thought about leaving this part from the demo. But then users would copy paste, possibly add `Options.inner` for focus purposes and then wonder why expandable does not work anymore

debois: Ok. Let's not also set 'inner` in the demo, then—users will think they need to.

debois: Ah, wait, I see. Hm.

debois: This is the best I can think of: 
- if user sets inner and expandable and they differ, log an error message on the console, use `inner`.
- if they set just one, use that one. 

?

vipentti: Unfortunately I don't think it's possible to easily detect setting the attributes due to `Options.id` and `Options.attribute <| Html.Attributes.id` get turned into `Html.Attribute` and we cannot access those values once set.

debois: Maybe we should change that; this situation is bound to come up again.

Maybe change option.id to apply only to things that have an id field in their Config, then delay turning that id into an attribute until some later point?

Don't know which point, though. It would suck if we have to apply it manually, and it wouldn't work with style then.

Maybe special-case id in the Property c m type? I normally wouldn't want that, but you could argue that id is like class in that it is (well, will be) a very common thing to both set, and inspect, like here?

vipentti:
Removed the `Options.inner` from the demo and renamed the parameter in the demo to `id-of-expandable-1` . Hopefully that makes it more clear that it should be the id of the element. 

Additionally prioritizing `expandableId`  over `Options.inner` because I'd prefer our components not breaking even if users provide "invalid" attributes (`id` in this case) but in general as well. 

This is one of those things that currently there is no other (as far as I know, without JS hacks) way to do this but maybe in v8 we could revisit this if we end up making MAJOR changes to `Options`.

vipentti:
@debois Do you think we can merge this even in its current state and then possibly revisit this for v8 ?

debois:
The API of this one will likely change once we get around to doing #208, but we're breaking stuff in v8 anyway, and it seems to be an important feature, so lets merge it.

PR #200, opened 2016-08-21
add space to ensure proper rendering of documentation for render

simonh1000:
I hope this helps. My editor seems to have stripped all trailing space from lines. The real edit is the addition of a libe break at 378 to ensure the example renders properly when turned into documentation


debois:
Very good, thanks!

PR #202, opened 2016-08-21
Add support for type' for Buttons.

vipentti:
This allows for setting the type of the button.

```elm
    Button.render ...
      [ Button.type' "button"
      ]
      [ ... ]
```

### DIFF
--- a/src/Material/Button.elm
+++ b/src/Material/Button.elm
@@ -114,7 +114,7 @@ type alias Config m =
   { ripple : Bool
   , onClick : Maybe (Attribute m)
   , disabled : Bool
-  , buttonType : Maybe String
+  , type' : Maybe String
   }
 
 
@@ -123,7 +123,7 @@ defaultConfig =
   { ripple = False
   , onClick = Nothing
   , disabled = False
-  , buttonType = Nothing
+  , type' = Nothing
   }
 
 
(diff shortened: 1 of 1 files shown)


src/Material/Button.elm
```diff
@@ -112,6 +114,7 @@ type alias Config m =
   { ripple : Bool 
   , onClick : Maybe (Attribute m)
   , disabled : Bool
+  , buttonType : Maybe String
```
debois: Why not just `type'`?

src/Material/Button.elm
```diff
@@ -235,13 +252,17 @@ view lift model config html =
         else 
           Nothing
       ] 
+
+    buttonType =
+      [ Maybe.map Html.Attributes.type' summary.config.buttonType ]
```
debois: Like the way you do in `LinkButton` better:

```elm
case summary.config.buttonType of
  Nothing -> [] 
  Just tp -> [ tp ]
```

debois:
Made a few comments, but otherwise, merge at will. 

Question of principle: Do we want the demo to exhibit or at least mention in text everything a component can do? Pro: It'd be great for users. Con: It'll be a ton of work for us, making we sure we have a demo of stuff like this, or at least mention it in text.

PR #205, opened 2016-08-23
Initial FAQ template

hakonrossebo:
I've just added some structure and text. I have not filled out all answers or linked this file from the readme.


debois:
I think I'd prefer not to duplicate things from the README, but otherwise, this looks great! If you could add a link from README to the commit and merge it, I'll fill in the missing questions. 

Thanks a bunch for doing this, it really helps!

hakonrossebo:
I added the link from README. I agree that we should not duplicate things from README, but we should maybe move some information from the README to the FAQ (or a Getting started guide) to simplify the front page information.

PR #206, opened 2016-08-24
Add support for id and inner attributes

### LINKED ISSUE #204: Add support for Id attribute
With the new [elm-lang/Dom](http://package.elm-lang.org/packages/elm-lang/dom/1.1.0/) package users can focus and blur elements based on the Id. 

For some use cases the following works
```elm
Options.attribute <| Html.Attributes.id "my-id"

-- For example with Textfield 
-- NOTE: Setting the attribute on style to set it to the actual input and not the container

, Textfield.style [ Options.attribute <| Html.Attributes.id "my-id" ]
```

But `Options.attribute` is restrictive for a reason as discussed in #192 . So adding `Options.id` might be a good idea. However, this has a problem that setting the id on a `elm-mdl` component might not result in what users might expect. 

For example settings `Options.id` on a textfield the expected result would be to have it affect the actual `input` and not the container, because to focus the textfield you need to set focus on the `input` and not the container. Similar problems might occur in other components where the component has several parts inside a container. 

This should be taken into consideration on how to best handle this particular matter. 

**NOTE** The problem would still exist, even if we made `Options.attribute` less restrictive.

vipentti:
Resolves #204

`Options.id` allows users to specify the `id` that can then be
used with the new DOM API.
`Options.inner` allows users to specify attributes on `inner`
elements of components that support it.

I went through the list by @debois in #204. Only components that I really found needing support for `Options.inner` were `Textfield` as well as `Toggles`, both of them have `input` elements that the users can now add attributes to with `Options.inner`. 

Most of the other components have users specify sections of content already and for those they can just use `Options.id` (such as Tabs and Grids). For other components there was really no clear `inner` element that would benefit from enabling this support.

As always looking for feedback and suggestions, especially if there is a component that could really use `Options.inner` support

### DIFF
--- a/src/Material/Textfield.elm
+++ b/src/Material/Textfield.elm
@@ -236,6 +236,8 @@ onFocus f =
 
 
 {-| Set properties on the actual `input` element in the Textfield.
+
+**Deprecated**. Use `Options.inner` instead. This value will disappear in 8.0.0.
 -}
 style : List (Options.Style m) -> Property m
 style =
--- a/src/Material/Textfield.elm
+++ b/src/Material/Textfield.elm
@@ -88,7 +88,7 @@ type Kind
   | Password
 
 
-type alias Config m = 
+type alias Config m =
   { labelText : Maybe String
   , labelFloat : Bool
   , error : Maybe String
@@ -99,7 +99,7 @@ type alias Config m =
   , cols : Maybe Int
   , autofocus : Bool
   , maxlength : Maybe Int
-  , style : List (Options.Style m)
+  , inner : List (Options.Style m)
   , listeners : List (Html.Attribute m)
   }
 
@@ -116,7 +116,7 @@ defaultConfig =
   , cols = Nothing
   , autofocus = False
   , maxlength = Nothing
-  , style = []
+  , inner = []
   , listeners = []
   }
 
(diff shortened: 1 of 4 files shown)


OvermindDL1:
What about `Menu`?  I've had need to attach things to both the outer container div as well as the button itself that brings up the menu.  Although this will not be needed if Menu is ever converted to be like ToolTip's where you 'attach' it, but currently it is not.  :-)

debois:
@OvermindDL1: @aforemny is looking into (some variant of) the attach idea, see  [here](https://github.com/debois/elm-mdl/issues/165#issuecomment-239429451).

src/Material/Textfield.elm
```diff
@@ -99,7 +99,7 @@ type alias Config m =
   , cols : Maybe Int
   , autofocus : Bool
   , maxlength : Maybe Int
-  , style : List (Options.Style m)
+  , inner : List (Options.Style m)
```
debois: We lost `style`, so this needs to be merged in master, not v7, yes?

src/Material/Textfield.elm
```diff
@@ -237,10 +237,9 @@ onFocus f =
 
 {-| Set properties on the actual `input` element in the Textfield.
 -}
```
debois: Mark as deprecated and remove it in master, i.e., 

```**Deprecated**. Use `inner` instead. This value will disappear in 8.0.0.```

debois:
I made some minor line notes, but I like this. Merge it when you've settled the v7/master question.

src/Material/Textfield.elm
vipentti: Textfield actually does not expose `Config` so this is a `MINOR` change according to `elm-package diff`

PR #207, opened 2016-08-24
Add support for Link Buttons

### LINKED ISSUE #201: Add LinkButton
This recently came up with trying to implement https://vipentti.github.io/elm-mdl-dashboard/ 's `View Source` button. 

In the actual MDL template the `View Source` button is technically Material button but it is a link-element (`a`).

Now I am uncertain if there is an actual need for this as you can somewhat work around this issue (opening a link) either by using Ports (required if you need to open the link in a new tab using window.open to emulate `target = "_blank"`) or for example `Navigation.newUrl`.

Basically this would be a `Button` in all ways except the `button` -element would be swapped to `a` and some additional attributes added: `href` and `target` at least.

**NOTE** This not strictly specified in either MDL and Material Spec. However, MDL allows this as it uses CSS classes and Material Spec does not take into account the type of the element, as long as it is a button which this technically still is, only the Html element is changed but the functionality is not.

vipentti:
Adds support for Link buttons to resolve #201 via 
```elm
    Button.render ...
      [ ...
      , Button.link
          [ Button.href "#some-link"
          , Button.target "_blank"
          ]
      ]
      [ ... ]
```
This allows for links that look and feel like buttons but perform link actions.
**NOTE** An empty list keeps the element as `button`

Additionally, using opaque `LinkProperty m` type to remove the possibility of introducing arbitrary `Html.Attributes` via `Button.link` (initial implementation allowed for Html.Attributes). These link properties could possibly be moved into a separate module for use in other modules if needed.

As always, open to feedback and suggestions

### DIFF
--- a/demo/Demo/Buttons.elm
+++ b/demo/Demo/Buttons.elm
@@ -242,7 +242,7 @@ view model =
     , Button.render Mdl [9, 0, 0, 1] model.mdl
         [ Button.ripple
         , Button.link
-            [ Html.Attributes.href "#buttons"
+            [ Button.href "#buttons"
             ]
         ]
         [ text "Basic Link" ]
@@ -252,7 +252,7 @@ view model =
       Button.render Mdl [9, 0, 0, 1] model.mdl
         [ Button.ripple
         , Button.link
-            [ Html.Attributes.href "#buttons"
+            [ Button.href "#buttons"
             ]
         ]
         [ text "Basic Link" ]
@@ -263,8 +263,8 @@ view model =
         , Button.colored
         , Button.raised
         , Button.link
-            [ Html.Attributes.href "https://github.com/debois/elm-mdl/"
-            , Html.Attributes.target "_blank"
+            [ Button.href "https://github.com/debois/elm-mdl/"
+            , Button.target "_blank"
             ]
         ]
         [ text "View Source" ]
@@ -275,8 +275,8 @@ view model =
         , Button.colored
         , Button.raised
         , Button.link
-            [ Html.Attributes.href "https://github.com/debois/elm-mdl/"
-            , Html.Attributes.target "_blank"
+            [ Button.href "https://github.com/debois/elm-mdl/"
+            , Button.target "_blank"
             ]
         ]
         [ text "View Source" ]
--- a/demo/Demo/Buttons.elm
+++ b/demo/Demo/Buttons.elm
@@ -214,14 +214,14 @@ viewButtons model =
 
 view : Model -> Html Msg
 view model = 
-  Page.body2 "Buttons" srcUrl intro references  
+  Page.body1' "Buttons" srcUrl intro references
     [ p [] 
         [ text """Various combinations of colors and button styles can be seen
                   below. Most buttons have animations; try clicking. Code for the
                   last clicked button appears below the buttons."""
         ]
     , Grid.grid [] (viewButtons model)
-    , p []
+    , Options.styled p [ Options.css "min-height" "250px" ]
         [ model.last 
            |> Maybe.map describe 
            |> Maybe.map (\str -> "Code for '" ++ str ++ "':")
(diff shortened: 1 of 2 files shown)


demo/Demo/Buttons.elm
```diff
@@ -214,22 +214,74 @@ viewButtons model =
...
         ]
-    ] 
+    ]
+    [ h4 [ id "link-buttons" ]
+        [ text "Link buttons"
+        ]
+
+    , p []
+        [ text """Link buttons are links that look and act like buttons.
+                However, they also support attributes for links such as 'href' and 'target'."""
```
debois: Drop "such as"—they support precisely those additional properties, right?

vipentti: You are correct! `Button.link` only supports the attributes that have been specified as LinkProperties

src/Material/Button.elm
```diff
@@ -71,13 +79,89 @@ import Parts exposing (Indexed, Index)
...
+
+
+{-| Specifies the URL of the page the link goes to
+-}
+href : String -> LinkProperty m
+href =
+  Html.Attributes.href >> Internal.attribute
+
+
+{-| Specifies where to open the linked document.
```
debois: Maybe add a link to 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a

for reader's unfamiliar with the target attribute

src/Material/Button.elm
```diff
@@ -71,13 +79,89 @@ import Parts exposing (Indexed, Index)
...
 import Material.Options as Options exposing (cs, when)
 import Material.Ripple as Ripple
+import Material.Options.Internal as Internal
 
 
 
--- MODEL
+{-| Opaque link property type
+-}
+type LinkProp = LinkProp
```
debois: Links occur also [in Layout](https://github.com/debois/elm-mdl/blob/v7/src/Material/Layout.elm#L481-L510). If we stick with `LinkProp` (see below), can we then refactor so that both use the same underlying implementation?

src/Material/Button.elm
```diff
@@ -105,13 +189,15 @@ update action =
...
 
+
 -- VIEW
 
 
 type alias Config m = 
   { ripple : Bool 
   , onClick : Maybe (Attribute m)
   , disabled : Bool
+  , linkOptions : List (LinkProperty m)
```
debois: Maybe just `link`?

src/Material/Button.elm
```diff
@@ -145,6 +232,26 @@ ripple =
...
 
 
+{-| Add link-attributes that turn the button from `button` into a `a`-element.
+This allows for links that look and feel like buttons but perform link actions.
+
+    Button.render ...
+      [ ...
+      , Button.link
+          [ Button.href "#some-link"
+          , Button.target "_blank"
```
debois: Does `target` have a reasonable default?

src/Material/Button.elm
```diff
@@ -145,6 +232,26 @@ ripple =
     (\options -> { options | ripple = True })
 
 
+{-| Add link-attributes that turn the button from `button` into a `a`-element.
+This allows for links that look and feel like buttons but perform link actions.
```
debois: Suggest rewrite:

```
{-|  Specify that the button should act as a link. This is convenient if you want pressing the button to re-direct the browser. 

Under the hood, this changes the implementation of button from `<button>` to `<a>`. 
-}

File: src/Material/Button.elm [hunk 7]
```diff
@@ -145,6 +232,26 @@ ripple =
...
+
+    Button.render ...
```
debois: Put in something reasonable for the dots to make the docs copy-paste'able.

src/Material/Button.elm
```diff
@@ -145,6 +232,26 @@ ripple =
...
+      [ ...
+      , Button.link
+          [ Button.href "#some-link"
+          , Button.target "_blank"
+          ]
+      ]
+      [ ... ]
+
+**NOTE** An empty list keeps the element as `button`
```
debois: I suggest making this not true by switching to `linkOptions : Maybe (List ...)`. I know its a corner-case that very few people will care about, but I think its better that the API contains as few surprises as possible.

src/Material/Button.elm
```diff
@@ -145,6 +232,26 @@ ripple =
     (\options -> { options | ripple = True })
 
 
+{-| Add link-attributes that turn the button from `button` into a `a`-element.
+This allows for links that look and feel like buttons but perform link actions.
+
```
debois: Can you still add an `onClick` handler? Maybe the docs should state whether you can and what will happen. Someone will try it at some point (e.g., if the link only does internal navigation in their SPA and they want something else to happen at the same time).

debois:
I like this. 

I'm a bit torn on all the attributes we wrap. I understand you do this to avoid people using `link` to override `onClick`, `onBlur` etc. Would we better off by implementing your #179 (custom events, dispatch) so users have a legitimate way to assign their own handlers, then leave them to use Attribute responsibly? I think the key question is if with what you've got, we've nailed _all_ the attributes people could conceivably want to add to a link. If we did, the present implementation  is fine. If instead this the seed of "attribute creep" (i.e., "we need also this one! and this one! and this one!", until we've wrapped all of `Html.Attributes`), maybe we should offer people the dangerous freedom of using `Html.Attributes` directly. 

Maybe the way to decide this is that someone takes a look at the possible attributes users might want to put on our current components, especially`Textfield` and `Button`, and see if a comprehensive list is reasonably short. Or at least not too long.

vipentti:
I'll hold off on fixing this until we figure out #208

vipentti:
I think this one now depends on either #213 or #179, if we want these in v7 then just generic attributes should be enough. However, we can also just push these directly to v8.

vipentti:
Closing this, will reopen once #213 or #179 have been merged. Generic attributes makes this trivial to implement.

PR #210, opened 2016-08-26
Elm-test framework and Travis CI speed improvements

hakonrossebo:
This adds a separate tests folder to elm-mdl. This does not affect elm-mdl in any way. I've also update the travis.yml file to use caching (reduces build time from 43 to 17 minutes) and use the test framework.

The current Travis CI build speed is 40-50 minutes. A couple of weeks ago, I tested the caching described in this [article](https://8thlight.com/blog/rob-looby/2016/04/07/caching-elm-builds-on-travis-ci.html), but with no luck. 

All other examples used [elm-test](https://github.com/elm-community/elm-test) as the last step in Travis CI, so I've implemented some basic tests using [elm-test](https://github.com/elm-community/elm-test). The Travis CI build is now down to 17 minutes - it's possible that it can be reduced more by caching the entire "elm-stuff" folder.

One primary reason I had to include elm-test now, is to improve the speed of Travis build.
Currently, there are no actual tests implemented. The Material components and Demo components are imported by the testrunner and only verifies that building those modules does not fail.

I've included a TESTING.md file with documentation on how to use testing in elm-mdl, but it's currently not required for running and developing elm-mdl.

The TESTING.md also references other projects using elm-test, to give some ideas on how to test elm-mdl. Testing can be helpful, but I'm not sure how elm-mdl Material components can be tested. Do you have any feedback or opinions on this?

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -1,6 +1,6 @@
 # Testing Elm-mdl
 
-### This is only relevant when developers and contributors to Elm-mdl
+### This is only relevant for developers and contributors to Elm-mdl
 
 Tests are located in a separate *"tests"* folder and does not affect elm-mdl.
 
--- a/TESTING.md
+++ b/TESTING.md
@@ -1,0 +1,25 @@
+# Testing Elm-mdl
+
+### This is only relevant when developers and contributors to Elm-mdl
+
+Tests are located in a separate *"tests"* folder and does not affect elm-mdl.
+
+Prerequisite: To be able to run tests from your terminal - install node-test-runner by running:
+> npm install -g elm-test
+
+Run all tests by executing this command in the root folder:
+>elm-test
+
+The travis.yml file has been configured to run elm-test when any branch is pushed to GitHub
+
+**NOTE**
+
+One primary reason to include elm-test now, is to improve the speed of Travis build.
+Currently, there are no actual tests implemented. The Material components and Demo components are imported by the testrunner and only verifies that building those modules does not fail.
+
+We should look at other projects to see how to implement tests in Elm-mdl.
+- [elm-test](https://github.com/elm-community/elm-test) - The elm-test package used
+- [elm-css](https://github.com/rtfeldman/elm-css) - A relevant project using elm-test.
+- [elm-combine](https://github.com/Bogdanp/elm-combine) - Another relevant project using elm-test.
+- [Caching elm builds on Travis](https://8thlight.com/blog/rob-looby/2016/04/07/caching-elm-builds-on-travis-ci.html)
+- [Article on testing in Elm](https://medium.com/@_rchaves_/testing-in-elm-93ad05ee1832#.tr0bpt18s) - Some parts are outdated.
--- a/tests/Tests.elm
+++ b/tests/Tests.elm
@@ -1,0 +1,63 @@
+module Tests exposing (..)
+
+import Test exposing (..)
+import Expect
+
+--Do not delete these imports as they are beeing compiled even when not beeing tested directly
+
+import Material
+import Material.Color as Color
+import Material.Layout as Layout
+import Material.Helpers exposing (pure, lift, map1st, map2nd)
+import Material.Options as Options exposing (css, when)
+import Material.Scheme as Scheme
+import Material.Icon as Icon
+import Material.Typography as Typography
+import Material.Menu as Menu
+
+import Demo.Page as Page
+import Demo.Buttons
+import Demo.Menus
+import Demo.Tables
+import Demo.Grid
+import Demo.Textfields
+import Demo.Snackbar
+import Demo.Badges
+import Demo.Elevation
+import Demo.Toggles
+import Demo.Loading
+import Demo.Layout
+import Demo.Footer
+import Demo.Tooltip
+import Demo.Tabs
+import Demo.Slider
+import Demo.Typography
+import Demo.Cards
+import Demo.Lists
+import Demo.Dialog
+import Demo.Chips
+
+
+
+mainSuite : Test
+mainSuite =
+    describe "Main Test Suite"
+        [ test "Sample test" <|
+            \() ->
+                Expect.equal (1) 1
+        ]
+
+materialButtonSuite : Test
+materialButtonSuite =
+    describe "Material Button Test Suite"
+        [ test "Sample Button test" <|
+            \() ->
+                Expect.equal (1) 1
+        ]
+
+
+all : Test
+all =
+    describe "elm-css"
+        [ mainSuite
+        ]
--- a/tests/Tests.elm
+++ b/tests/Tests.elm
@@ -47,17 +47,8 @@ mainSuite =
                 Expect.equal (1) 1
         ]
 
-materialButtonSuite : Test
-materialButtonSuite =
-    describe "Material Button Test Suite"
-        [ test "Sample Button test" <|
-            \() ->
-                Expect.equal (1) 1
-        ]
-
-
 all : Test
 all =
-    describe "elm-css"
+    describe "elm-mdl"
         [ mainSuite
         ]
--- a/tests/elm-package.json
+++ b/tests/elm-package.json
@@ -1,0 +1,26 @@
+{
+    "version": "1.0.0",
+    "summary": "Test project for Elm-mdl",
+    "repository": "https://github.com/debois/elm-mdl.git",
+    "license": "BSD-3-Clause",
+    "source-directories": [
+        ".",
+        "../examples",
+        "../demo",
+        "../src"
+        ],
+    "exposed-modules": [],
+    "dependencies": {
+        "elm-lang/core": "4.0.0 <= v < 5.0.0",
+        "elm-community/elm-test": "2.0.0 <= v < 3.0.0",
+        "rtfeldman/node-test-runner": "2.0.0 <= v < 3.0.0",
+        "debois/elm-dom": "1.2.0 <= v < 2.0.0",
+        "debois/elm-parts": "4.0.0 <= v < 5.0.0",
+        "elm-lang/html": "1.1.0 <= v < 2.0.0",
+        "elm-lang/mouse": "1.0.0 <= v < 2.0.0",
+        "elm-lang/window": "1.0.0 <= v < 2.0.0",
+        "elm-lang/navigation": "1.0.0 <= v < 2.0.0",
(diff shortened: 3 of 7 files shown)


OvermindDL1:
Curious question.  I just ran the .travis.yml locally and everything completed successfully in <120s (and this is not a fast system, although it does have an SSD).  Why would it take so many minutes on travis?

hakonrossebo:
Not sure, but I think Travis CI has a limited set of resources for open source projects, as discussed [here on SO](http://stackoverflow.com/questions/29567063/how-does-travis-ci-org-throttle-builds). Most workarounds seems to focus on caching and running different tests in parallell.

debois:
Occasionally, my build will break locally and I have to clear out elm-stuff to the build system back into working order. Is there an easy way to manually clear the elm-stuff in the Travis cache?

debois:
I'm not sure how to do testing of components. Usually, if it builds, it works. The exception is browser incompatibilities (we love you, Firefox), but I'm not sure how to test for that.

We could do unit tests of important infrastructure like Options, but we don't touch that very often, so I don't know if it'd be worth it.

hakonrossebo:
The Travis cache can easily be deleted in the Travis CI user interface under "more options - cache".

On testing and test framework - The main reason I added it now was to be able to have something running and exiting with 0/1 in the scripts tag in travis.yml - I've tried other options with no luck.

I guess testing would be beneficial if we could be able to check/inspect the html returned from elm-mdl functions - then we could write unit tests/TDD to if verify if css classes has been set or not.

Anyway - For now It's only a workaround to help speeding up Travis CI builds. It's ok for me if we remove this PR now - we can always use this code later if we find use for it.

vipentti:
I have been entertaining the idea of possibly trying https://github.com/lorenzo/elm-webdriver to do just that. Inspecting the generated HTML based on events and such, to verify proper classes have been set and such

hakonrossebo:
Yes, I've used Nightwatch.js earlier and I think elm-webdriver could help us in not breaking any existing implementations going forward.

TESTING.md
```diff
@@ -0,0 +1,25 @@
...
+>elm-test
+
+The travis.yml file has been configured to run elm-test when any branch is pushed to GitHub
+
+**NOTE**
+
+One primary reason to include elm-test now, is to improve the speed of Travis build.
+Currently, there are no actual tests implemented. The Material components and Demo components are imported by the testrunner and only verifies that building those modules does not fail.
+
+We should look at other projects to see how to implement tests in Elm-mdl.
```
debois: Very good.

Maybe open an issue that we need test and move these links there?

tests/elm-package.json
```diff
@@ -0,0 +1,26 @@
+{
+    "version": "1.0.0",
+    "summary": "Test project for Elm-mdl",
+    "repository": "https://github.com/debois/elm-mdl.git",
+    "license": "BSD-3-Clause",
```
debois: "Apache License, Version 2"

debois:
Would it be possible to add the "check if you're on the right branch" script we discussed on the elm-slack? I'd really like that. 

Either way, I'm good with merging this.

TESTING.md
hakonrossebo: I can do that, to confirm - should I create an issue that we need testing, and then add the information from the TESTING.md into the issue?

hakonrossebo:
I updated the license. I think I'll be able to add those PR/branch/package-diff scripts to Travis - just been too busy at work.

OvermindDL1:
You might want to comment on the experiences about this at https://github.com/elm-lang/elm-compiler/issues/1473 as there is an active conversation going on.

jvoigtlaender:
1. In addition to caching `elm-stuff/build-artifacts` it makes sense to also cache `elm-stuff/packages`, because that will prevent repeated downloading of the same package source bundles (dependencies used) again and again.
2. The thing about having to clean out the cache from time to time in order to delete `elm-stuff` that might otherwise be corrupted, might be solved by https://github.com/elm-lang/elm-make/commit/46ec85c0e8dc5f3427bde6e4f212b4f367ecbd07.

jvoigtlaender:
@robertjlooby, since your article on Elm build caching is used above (and also elsewhere I've seen) as source for how to do caching on Travis: I think it is a fact that adding `elm-stuff/packages` (but *not* `elm-stuff/exact-dependencies.json`) to the to-be-cached directories is always safe. It does not have the kind of side effects your article warns against. Do you agree? If you have contradictory experience, my suggestion 1. in the previous comment is wrong, and it would be good to know. If you do agree, though, then maybe an update to your article would be reasonable, as it informs a too pessimistic caching strategy.

debois:
@hakonrossebye: if you have time, it looks like the thread @overmindDL1 referenced has a pretty effective fix for reducing CI-build times.

robertjlooby:
@jvoigtlaender There aren't any downsides to caching `elm-stuff/packages` as well, but it just doesn't give a huge overall performance boost like caching `elm-stuff/build-artifacts` does. I discussed it some here https://github.com/elm-community/elm-test/issues/42#issuecomment-240619618.

@OvermindDL1's comment was the first I've seen of that thread. It looks like the issue has finally made it on Evan's radar so there may be a cleaner compiler fix coming.

I might update the blog with info about caching packages, but I definitely will whenever there's a flag/fix in the compiler.

hakonrossebo:
@debois, @OvermindDL1 - I'v been following this thread the last days, but I have not had the time to look into it. It looks to me that the workaround is to run `sysconfcpus -n 1 elm-make`. I'm on a windows machine and not sure how to proceed with building sysconfcpus on my machine and then also set this as a dependency to be downloaded and built in Travis CI. I can ask @obmarg for more details on how to do this if we want to proceed - speeding up build time would be great.

OvermindDL1:
You can actually build sysconfcpus on travis itself, then cache the output and only recompile next time if it does not already exist.  :-)

obmarg:
@hakonrossebo if it's of interest, [here's an outtake of my circle.yml for building libsysconfcpus](https://gist.github.com/obmarg/89eb0c6cbe3826d7859683ad987db691)  Think it should be easy enough to convert to travis.  Travis container builds have no sudo support, so you may need to work around that by just running libsysconfcpus from it's build directory.

OvermindDL1:
Travis has sudo support if you explicitly list it as needed.  However it should not be required and you can run it from its build directory (no need to `make install` it at all, just build).

debois:
@hakonrossebo: Speeding up Travis build times would be very nice indeed. However, as @jvoigtlaender says, it seems a patch for elm-make will come eventually. 

So, if you have the time, this'd be a great thing to have _now_; but it might come in 0.17.2 or 0.18; so we could also hold off for a few months if we want to be conservative with our time?

jvoigtlaender:
I've made a pull request against @hakonrossebo's branch that implements this "enforce correct cpu number on Travis" optimization: https://github.com/hakonrossebo/elm-mdl/pull/1

It brought down the time for compiling the 60+ modules of dependencies from something like 20 minutes to 1 minute.

The Travis build is still failing there, but that's not caused by this change, instead it was already failing before, with an unrelated error.

As another note: Waiting for a fix to `elm-make` is not enough. You will also then have to wait for `elm-test` to be changed in such a way that you can pass a flag to `elm-test` which will cause `elm-test` to make use of the newly introduced flag of `elm-make`...

So I suggest taking this fix now.

hakonrossebo:
This is great - thanks for helping out. Saved me from spending time in unknown territory. I also agree that we should use this fix now instead of waiting. 

I'm not really sure why the build is failing - I updated elm-test locally and updated the elm-test package dependency to elm 0.17.1 earlier today, but it's still failing. The Travis CI log reports `elm-community/elm-test/2.1.0/Fuzz.elmo: openFile: does not exist (No such file or directory)` as the reason. I've deleted the Travis cache so you could possibly try to update the PR again to see if it succeeds.

jvoigtlaender:
Did so, and the build succeeds now: https://travis-ci.org/hakonrossebo/elm-mdl/builds/158548257.

hakonrossebo:
Excellent, @debois - should we just merge this PR then?

jvoigtlaender:
@hakonrossebo, you would first have to merge my PR into your branch, before it would become visible here. (But maybe that's what you mean by your preceding question.)

tests/Tests.elm
```diff
@@ -0,0 +1,63 @@
...
+    describe "Material Button Test Suite"
+        [ test "Sample Button test" <|
+            \() ->
+                Expect.equal (1) 1
+        ]
+
+
+all : Test
+all =
+    describe "elm-css"
```
debois: Maybe "elm-mdl"? :)

tests/Tests.elm
```diff
@@ -0,0 +1,63 @@
+module Tests exposing (..)
+
+import Test exposing (..)
+import Expect
+
+--Do not delete these imports as they are beeing compiled even when not beeing tested directly
```
debois: "being" (twice)

TESTING.md
```diff
@@ -0,0 +1,25 @@
+# Testing Elm-mdl
+
+### This is only relevant when developers and contributors to Elm-mdl
```
debois: when -> for

tests/Tests.elm
```diff
@@ -0,0 +1,63 @@
...
+
+mainSuite : Test
+mainSuite =
+    describe "Main Test Suite"
+        [ test "Sample test" <|
+            \() ->
+                Expect.equal (1) 1
+        ]
+
+materialButtonSuite : Test
```
debois: Maybe loose this as its currently not relevant?

debois:
I made a few nit-pick line comments above, but otherwise, yes, please merge this—it'll be _so_ good to have decent travis build times. 

As far as I can tell, we still don't check that the commit is on the right branch (right?), but let's save that for a future PR.

tests/Tests.elm
hakonrossebo: Thougth so myself, but it failed afterwards - Almost same name on definition and import.

jvoigtlaender: I think @debois was talking about the type "beeing" (twice) in the comment.

hakonrossebo:
Thanks to everyone that helped out with this!

I've updated the code from @debis last comments. 

The "commit is on the right branch" will be a separate PR - hopefully soon. 

I will also create a separate issue that we need testing, and include some of the testing links there.

PR #213, opened 2016-08-27
Initial implementation of generic attribute support

### LINKED ISSUE #208: Support attributes in options
Options currently support `Html.Attribute.attribute` only on `Style m` options. Other components laboriously wrap parts of `Html.Attribute` in `Options` to allow users customisation of various sorts. We need the latter to protect ourselves from users breaking elm-mdl by overriding important attributes, usually `on "..."`.

Wrapping `Html.Attribute` is suspect because (a) it seems wasteful and inelegant the more attributes we add, and (b) we end up in situations where distinct components need the same attribute wrapped, and currently have no good solution for that. See #201.

I'd like to discuss whether the current approach of wrapping `Html.Attribute` values is the way forward, or if we should aim for something more general. 

For starters, let's make a list of attributes currently supported and missing.

vipentti:
This is a initial implementation of generic attribute support
to enable arbitrary `Html.Attributes` to resolve #208 . This implementation
tries to minimize the likelihood of overriding internally used
attributes while still maintaining `MINOR` change.

This implementation may be revisited in the next MAJOR version
to update the implementation.

Add namespaced qualifiers to `Internal.attribute`,
this way internal attributes are easy to search for as well separate
them from `Options.attribute`.

As always, open to feed back and suggestions!

### DIFF
--- a/src/Material/Card.elm
+++ b/src/Material/Card.elm
@@ -50,7 +50,7 @@ import Html exposing (..)
 import Html.Attributes
 
 import Material.Options as Options exposing (Style, cs, css)
-import Material.Options.Internal exposing (attribute)
+import Material.Options.Internal as Internal
 
 
 {-| Separate given content block from others by adding a thin border.
@@ -157,7 +157,7 @@ can't use `Html.Events.onWithOptions` because we can't construct a
 -}
 stopClick : Style a
 stopClick = 
-  attribute <| 
+  Internal.attribute <|
     Html.Attributes.attribute
       "onclick"
       "var event = arguments[0] || window.event; event.stopPropagation();"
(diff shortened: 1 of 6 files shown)


debois:
Sorry about the late reply. I like this! You closed it again?

vipentti:
I guess I can update it if we want it in the current v7

vipentti:
This should reduce the likelihood of overriding internal attributes by separating them.

debois:
Just to be clear: #179 contains this already, yes? In that case, do we want this now, or can it wait for v8?

vipentti:
Yes, #179 contains generic attribute support as well. We may wait or we may use this for now. Either way works for me. I think it depends if this is needed now and if v8 is some time a way still.

debois:
Ok, let's do it. It keeps coming up.

PR #216, opened 2016-08-30
Improve Options.collect* performance

debois:


vipentti:
Looks good! As far as I'm concerned this can be merged.

debois:
Thanks—could you do it? I want to be all principled and not do it myself :)

debois:
:)

PR #225, opened 2016-09-06
Correct project description in USERS.md

IwalkAlone:
Accidentally noticed that my project is in USERS.md. I was confused why I was getting traffic on it all of a sudden :)

Here's a correct description for my project - the previous one was leftover from a starter kit I forked. Wouldn't want people to come there thinking it's a webpack boilerplate :) (that would be https://github.com/moarwick/elm-webpack-starter, which is awesome)


debois:
Great, thanks!

PR #246, opened 2016-10-20
Support input with type ‘email’

rmies:


debois:
Thanks! Apologies for the long wait.

PR #252, opened 2016-11-14
Fix source URL typo on Cards page

MoonlightOwl:


debois:
Thanks! Apologies for the long wait.

Issue #256, opened 2016-12-04
Cards.elm demo, broken source code link.


swojtasiak:
Link to source code is broken in Cards.elm example. It points to Card.elm instead of Cards.elm.

swojtasiak:
https://github.com/debois/elm-mdl/pull/257

PR #257, opened 2016-12-04
#256 broken source code link fixed

swojtasiak:
Broken reference to Cards.elm example source code fixed.


debois:
Thanks! Apologies for the wait!

PR #263, opened 2016-12-29
added Options.node to enable adding any type of node to an elm-mdl node

MarkNijhof:
I want to embed some SVG in the button which I think I cannot do currently:
```
Button.render Mdl
                [ 0 ]
                model.mdl
                [ Button.icon
                ]
                  [ Options.node "svg" [ css "width" "24px", css "height" "24px", viewBox "0 0 24 24"]
                                                      [ Options.node "path" [Options.attribute fill "#000000",  Options.attribute d="M16,12V4H17V2H7V4H8V12L6,14V16H11.2V22H12.8V16H18V14L16,12Z"
                                                      ]
                ]
```
Looking at the elm-mdl code again, I guess I can do it with `Options.styled Html.node "svg" [] []` but that is not very discoverable :)

### DIFF
--- a/examples/Counter.elm
+++ b/examples/Counter.elm
@@ -103,12 +103,12 @@ view model =
             [ Options.onClick Increase
             , css "margin" "0 24px"
             ]
-            [ text "Increase" ]
+            [ Options.node "div" [] [ text "Increase" ] ]
         , Button.render Mdl
             [ 1 ]
             model.mdl
             [ Options.onClick Reset ]
-            [ text "Reset" ]
+            [ Options.node "div" [] [ text "Reset" ] ]
         ]
         |> Material.Scheme.top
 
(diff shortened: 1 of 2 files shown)


debois:
In this particular case, couldn't you just forget about `Options` altogether and do:

    Button.render Mdl [ 0 ] model.mdl
      [ Button.icon ]
      [ Html.node "svg" 
          [ class "width" "24px", class "height" "24px", viewBox "0 0 24 24" ]
          [ Html.node "path" 
              [ fill "#000000"
              , d "M16,12V4H17V2H7V4H8V12L6,14V16H11.2V22H12.8V16H18V14L16,12Z"
              ]
              []
          ]
      ]


I like your general `Option.node`, but I'm not sure when it would be needed? `Option.div` and friends already cover the common case, and there's `Material.styled` for the rest.

MarkNijhof:
lol that worked, I just assumed that since this was inside a Button I had
to provide elm-mdl nodes...

debois:
I'm closing this for now, in the absence of a clear use-case. Feel free to re-open if one comes to mind!

Issue #260, opened 2016-12-28
Roadmap


debois:
[Continuing discussion from #253.]

Google won't continue development on MDL, recommending instead a switch to MDC (can someone furnish a link corroborating this statement?)

This leaves the following challenges for elm-mdl:

1. Short-term, elm-mdl needs textfield with autocomplete and textfield with dropdown. Since these are not going to come from Google, I suppose we make some ourselves. Does anyone want to take a first stab at either? (It might be possible to repurpose code from Menu.) I'd be very happy to assist from the sidelines. 

2. Long-term, where does elm-mdl go in the absence of MDL v2? The obvious choice seem to be keep the API, port over the implementation to MDC. I don't understand what's up with web components and elm (last I tried, I couldn't get the former to work with virtual dom), and I'm unsure of the ramifications. Insights would be greatly appreciated.

tolgap:
The link corroborating that statement is at `google/material-design-lite` repo's README: https://github.com/google/material-design-lite#limited-support

> Limited support
> 
> Material Design Lite is now in limited support, with development having moved to the Material Components for the web repository.
> 
> No further development is taking place in MDL by the core team, but we are happy to review PRs, fix critical bugs and push out new releases. No breaking changes will be accepted.

debois:
Well, duh.

brandonmp:
@debois - I just started playing with Elm, but one thing I've read a few times is that `elm-mdl`'s approach to eliminating boilerplate is at odds with Elm's author's vision language norms & patterns.  

This [thread on google groups](https://groups.google.com/d/msg/elm-discuss/dC0htGQOssY/xhHMjWd9EAAJ) implied that this difference of opinion might be terminal for `elm-mdl`, as there's some non-zero chance that the future Elm API could introduce breaking changes. 

Seeing this roadmap of course implies that you don't agree. So can you opine briefly on how you view your boiler-plate reduction approach wrt core Elm? Do you expect to change the `elm-mdl` API in the future to comport with whatever patterns Elm decides upon? 

I'm just learning Elm, so forgive me if there's obvious background information I'm lacking here--just trying to understand what the UI Kit ecosystem is like in Elm before committing, and `elm-mdl` is clearly one of the top libraries in that arena.

nsb:
@brandonmp This should answer your question https://github.com/debois/elm-mdl/issues/253#issuecomment-269220106

debois:
[...]

PR #273, opened 2017-01-08
Textfield with dropdown component

aforemny:
Hi all,
in response to #260 I would like to share a textfield with dropdown implementation that I have been working on.

It does not integrate with menus, so there is some code duplication, but I am still uncertain about an implementation in Elm so that I do not want to create dependencies between the two modules. It adds sassc as a dependency to the Makefile because we have to generate select.css from styles/select.scss which in turn depends on material-design-lite which this does not track.

The component's demo is [here](http://aforemny.github.io/elm-mdl/#select).

The implementation uses some JS hacks that I would like to get rid of, but I do not have an adequate replacement for them in Elm, see. `Note:` in Material.Select.

Important to note is that I do not close the menu on blur, but rather on selecting a value or clicking outside. This has to do with the fact that (last time I checked) the component's blur event fires before a global mouse click event, so the menu is closing before you are able to select something. If you want to close a select component whenever another component receives focus, you'll have to call `Material.Select.closeAll` (not in the commit yet) whenever any component receives focus.

I chose this way, because re-ordering the relevant events in Elm requires me to use `Helpers.delay` and I do not like that this adds a constant delay to the component that is not reliable under load.

Should I polish this up and merge it? Any feedback welcome!


gotcha:
This is nice work.
Is it a design choice of MDC to make it so that typing first letters does not change the selection ?

debois:
I checked out only the demo so far, but I think this is awesome! We should definitely merge this once the kinks have been worked out. I've set some time aside to take a proper look monday; I'll get back to you then.

debois:
@hakonrossebo: I also see this one as not passing Travis. Can you help?

aforemny:
@debois Super! I appreciate it!

Travis CI seems to fail because I did not type annotate every function yet, in particular Material.Helpers.{fst,snd} which I added there ad hoc (where did fst, snd go by the way?) However, I am surprised that it seems to affect the build state of the v8 branch.

@gotcha: I have no idea. I just found out now, that [Polymer's dropdown demo](https://www.webcomponents.org/element/PolymerElements/paper-dropdown-menu/demo/demo/index.html) has that feature, I just did not notice it before. Still, I think this feature should be handled in user-code (as opposed to library code), because you can pass `List (Html m)` to your menu (as opposed to `List String`). Would you appreciate an example in the [demo](http://aforemny.github.io/elm-mdl/#select)? Would you want to write one (after this has been merged)?

aforemny:
I managed to get in there an implementation of `closeAll`. The code example is not very nice, and we could make the type nicer probably by including support in `Material.Commponent.Msg`.

I think the close on blur behavior is not intuitive, the docs should make that clear now:

> The select component closes itself on pressing TAB. It does not close
automatically if the user focuses another component by other means, ie. clicking.

> If you want select components to close if another element receives focus, you
will have to listen to that element's focus event and call
`Material.Select.closeAll` from your program.

src/Material/Helpers.elm
```diff
@@ -86,6 +88,20 @@ blurOn evt =
 -- TUPLES
 
 
+{-| Return the first element of a tuple.
+-}
+fst : (a, b) -> a
```
debois: In 0.18, these are `Tuple.first` and `Tuple.second`.

debois:
Hi @aforemny, I like this; it seems very reasonable to repurpose Menu functionality. Things we need to overcome—you probably thought of all these already, but let me write them down anyway:

- We have to find some way to factor out the large common ground between Select and Menu. Maybe a good way to do this is to try to factor out `Msg` parts and `update` function parts into a shared sub-message and a shared sub-update for starters?
- We need keyboard entry (e.g., hit 'a' to select 'allosaurus'.). This would be good to have in Menu too. 
- We have to get rid of the flashing when you click the textfield while the menu is open. 
- Menu currently doesn't close on ESC.
- Menu width should preferably match the width of the underlying textfield
- Would it be possible to display the menu below rather than over the textfield?

I definitely think you should go forward with this. It would be great to incorporate thinking about autocomplete, which will need some of the same functionality (specifically the menu).

debois:
PS. I didn't do a detailed line-by-line yet, since we're obviously still in an early state here.

aforemny:
I started refactoring common parts from Menu and Select into Dropdown. That was well worth it! I still have to refactor `Config`, add docs, etc. But I updated the PR with what I have so far.

> We need keyboard entry (e.g., hit 'a' to select 'allosaurus'.). This would be good to have in Menu too.

I agree. I will probably not attempt it within this PR, though.

> We have to get rid of the flashing when you click the textfield while the menu is open.

What do you mean here? Do you mean to remove the ripple effect on the input when the menu is open? If so, [Polymer](https://www.webcomponents.org/element/PolymerElements/paper-dropdown-menu/demo/demo/index.html) also has that effect (except the menu closes there in addition). I am considering to close the Menu as well to be closer to Polymer's implementation.

> Menu width should preferably match the width of the underlying textfield. Would it be possible to display the menu below rather than over the textfield?

That differs from Polymer's implementation. Still, I would go forward implementing it like that and add styling options for that later. With Select/Autocomplete, I have three additional alignments in mind that I would want to support. (Alignment is currently a bit broken in the menu demo.) I agree that width should be configurable.

aforemny:
I refactored the Config part, so the API is mostly done.

There are still improvements to be made to Material.Dropdown component, and I would rather not make that module public for now. How do I specify that in elm-package.json?

One issue that I am having with the Dropdown implementation is that the Menu relies on a button and because of quirks mode, the button raises a click event when it is focused and you hit space. This is raised both on the document (via Mouse subscriptions) and on the button itself (via on "click"), so I have to prevent two events when opening the menu with space bar. Somehow I have to prevent only one event while closing (with space bar). I would like to get rid of that in Menu (rather than in Dropdown), but I don't know how to.

I would like to check how Polymer does that with its iron-dropdown and paper-dropdown-menu, but for this PR I would be OK with leaving it as is, and only improve the Select styling and documentation some more.

If you find the time I would appreciate it if you could share your thoughts on Menu/Select - Dropdown inter-op, in particular Menu/Select's Config and how they are passed down to Button/Textfield's view and their update implementations.

edit: I have to expose Textfield.Config for that.

aforemny:
Closing in favor of #282.

PR #280, opened 2017-02-01
Swap numeric and non-numeric

jhntrifork:
Hi,

Great library. I noticed that "--non-numeric" is used when "numeric" is enabled and not the other way around.

Cheers,
Jon


aforemny:
Cherry-picked it in new-v8 (78891390c754c6621d8630dd30125f5797291c0b). Thanks!

PR #289, opened 2017-02-24
Fix textfield demos to use Textfield.value to manage is-dirty

vipentti:
Currently `Textfield` does not properly update its internal is-dirty
state when using user provided `Options.onInput`. Workaround
is to call `Textfield.value` with the input.

This would cause floating labels to not float properly when unfocused.


aforemny:
This should be fixed with #333.

Issue #240, opened 2016-10-03
Tooltip doesn't disappear


slegrand45:
Hello,

The following demo gives an example of the problem: https://slegrand45.github.io/test.site/www/

When you click on the "add" icon, the tooltip is correctly displayed and the page content changes. Then, when you click on the "list" icon, the previous content is displayed again but also the tooltip. Despite the fact that the mouse is not on the "add" icon.

Is there a workaround to avoid this behavior? I tested with Firefox 48 and Chrome 52.0.2743.116, both on Linux. The source code of the demo is available here: https://github.com/slegrand45/test.site/blob/master/Main.elm

Thanks.

vipentti:
Thanks for reporting this! Can verify this on Windows 10 and Chrome  54.0.2840.41 beta-m (64-bit). I'll take a closer look and see if I can find the cause, I believe it might be something to do with Tooltip currently using 'mouseleave' which might not be triggered properly when the click event causes the update to be called.

rofrol:
Bug can be close. Tested with elm-mdl-8.0.0.

Example code https://github.com/rofrol/elm-mdl-bug-240

slegrand45:
By the way, it's also visible on the [Tooltips demo page](https://debois.github.io/elm-mdl/#tooltips) but only with the button example (the "Click Me!" button).

kittykatattack:
Hi everyone, I just want to confirm that I'm experiencing the same problem.

debois:
Can someone check out if this is a known bug from upstream?

vipentti:
This is a bug due to how the events are handled currently

aforemny:
There are two issues being discussed here. The original issue relates to the fact that the tooltip does not disappear when you remove an element from DOM: apparently it does not fire the mouseleave event which is required for the Tooltip to hide correctly. I do not know any way around this.

The other issue that mentions the Tooltips demo page should be resolved in current new-v8. You can verify if it has been resolved in the [wip demo](https://debois.github.io/elm-mdl/wip.html#tooltips).

PR #290, opened 2017-02-24
Fix Tooltip when attaching to another mdl component

vipentti:
Use `Options.on` instead of `Html.Events.on` to attach
the event listener. This at least temporarily fixes the issue
where hovering over an MDL component such as `Button` with a tooltip
would cause the tooltip to stay visible.

NOTE: This might have unintended consequences when using `Tooltip`
with an MDL component with a ripple effect which also uses the
`mouseleave` event.

This at least temporarily resolves the `Click me` case from #240

### DIFF
--- a/src/Material/Tooltip.elm
+++ b/src/Material/Tooltip.elm
@@ -474,9 +474,9 @@ render =
 onMouseEnter :
     (Component.Msg button textfield menu snackbar toggles Msg tabs dispatch -> m)
     -> Component.Index
-    -> Attribute m
+    -> Options.Property c m
 onMouseEnter lift idx =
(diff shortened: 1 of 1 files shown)


thotypous (approved):
It works. I was trying to attach a Tooltip to some text or icons inside a Card, and this PR fixed it.

achauve:
Anyone to merge this?

aforemny:
I cherry-picked this in new-v8 (f2c6628f63a6ed28113ef29be597b8cd13a3dc72). Thanks!

Do you have a demo for the unintended side-effects with ripple? Otherwise, I will leave that potential problem to worry about later. :o

PR #292, opened 2017-03-06
Fix regression on the handling of deleteClick

thotypous:
Commit 98f4284907e9fac removed (apparently accidentally) some code which used to set up the Chip model's `deleteClick` attribute. After 98f4284907e9fac, `deleteClick` will never be set. It will always be equal to `Nothing`.

The following snippet currently just overrides `Log "onClick"` with `Log "deleteClick"`. It is thus impossible to set up both `onClick` and `deleteClick` in the same Chip correctly. Please note that `deleteLink` was not affected by this issue. 

```elm
Chip.span
    [ Chip.deleteIcon "cancel"
    , Options.onClick (Log "onClick")
    , Chip.deleteClick (Log "deleteClick")
    ]
    [ Chip.content []
        [ text "chip" ]
    ]
```

By partially reverting the change as per this pull request, the issue is fixed.

### DIFF
--- a/src/Material/Chip.elm
+++ b/src/Material/Chip.elm
@@ -40,6 +40,7 @@ for a live demo.
 
 import Html exposing (Attribute, Html)
 import Html.Attributes
+import Html.Events
 import Material.Options as Options exposing (cs)
 import Material.Icon as Icon
 import Material.Options.Internal as Internal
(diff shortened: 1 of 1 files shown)


aforemny:
Cherry-picked it in new-v8 (08a5bb1b7cfc618937244b9efa27a0963cab2248). Thanks a bunch!

PR #294, opened 2017-03-11
Fix minifab buttons not applying fab-class

vipentti:
Minifab buttons also need mdl-button--fab class.


aforemny:
Cherry-picked it in new-v8 (26dc3ab71f64b582211fab76415a403c27601137). Thanks! :)

PR #316, opened 2017-05-25
Updated docs -> Tooltip.default is not exposed

ghost:
Updated docs -> Tooltip.default is not exposed (nor exists)


aforemny:
Cherry-picked it in new-v8 (e6321cb515a2639890ef40449685172bd2ea2935). Thanks!

PR #322, opened 2017-06-10
Fix slider thumb in MS Edge

mbaileys:
This fixes issue #1625 by preventing the thumb from being clipped. The issue exists in [https://github.com/google/material-design-lite/blob/mdl-1.x/src/slider/_slider.scss](Google MDL).


aforemny:
Cherry-picked it in new-v8 (c37e1fdb95d1b203d1154ab75e5ecd07a6e29ed7). Thanks!

PR #349, opened 2017-08-11
Force chips to span tag when they contains actions

### LINKED ISSUE #332: Chips deleteClick is broken in Firefox
In the [new-v8 demo](https://debois.github.io/elm-mdl/wip.html), Chips are broken in Firefox (interactive demo, cannot delete chip). It works in Chrome.

alvivi:
HTML buttons can only contain phrasing elements. Some browsers (like
Firefox) does not support interactive elements inside a button. This
commit force chips to use a span element when they contain an action,
even when they are rendered to a button.

Fix #332.

### DIFF
--- a/src/Material/Chip.elm
+++ b/src/Material/Chip.elm
@@ -120,6 +120,7 @@ deleteClick =
                 }
            )
 
+
 type alias Priority =
     Int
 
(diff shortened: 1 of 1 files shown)


alvivi:
Just for reference: https://html.spec.whatwg.org/multipage/form-elements.html#the-button-element

The [mdl](https://getmdl.io/components/index.html#chips-section) documentation is confusing. The last example is wrong, but the deletable chip that works in Firefox is actually a `span`.

aforemny:
Merged it manually into v9. Thanks a bunch! :)

alvivi:
@aforemny should I start future pull requests on v9 branch even their issue is targeted to v8?

aforemny:
@alvivi Please open them on v9.

PR #350, opened 2017-08-19
Fix : hide Grid.All

ashishnegi:
Add all elm-mdl--hide-* classes when hide with Grid.All devices is used.

Current : When we use `Grid.hide Grid.All` we don't add any hide class for the device.
Expected : We should add hide class for all devices.

Use case :

in elm grid : 
I may add `hide Style` on the grid Cell based on some logic. E.g. on press state of a help-button, we want to show a help message. so, to logic of showing help message used `Grid.hide Grid.All` with a condition.

```
Grid.cell [ if model.showHelpMsg 
               then Grid.size Grid.All 12 
               else Grid.hide Grid.All 
                ] 
               [ .. view .. ]
```

### DIFF
--- a/src/Material/Grid.elm
+++ b/src/Material/Grid.elm
@@ -42,7 +42,7 @@ Example use:
 
     import Material.Grid exposing (grid, cell, size, Device(..))
 
-    top : (Html a)
+    top : Html a
     top =
       grid []
         [ cell [ size All 4 ]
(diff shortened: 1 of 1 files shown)


aforemny:
Merged in 21b17feb37baf413b5eb5f31ef53e37ae7f90c98. Thanks a bunch!

PR #359, opened 2017-12-06
Remove extra - from class name

### LINKED ISSUE #356: Hide function not working
Hello,

I found a bug on hide function inside the Grid module.
It code is:

``` elm
hide : Device -> Style a
hide device =
    cs <|
        case device of
            All ->
                ""
            _ ->
                "mdl-cell--hide-" ++ suffix device
```

the last hyphen in `"mdl-cell--hide-` is unecessary because `suffix device` alredy has a hyphen at it start.

katjam:
As noted by @bruunoromero the suffix device has a hyphen at the start.

The hide class should only contain one - like `mdl-cell--hide-phone`.

Fix #356


katjam:
Closing. The library is no longer being updated.

